PR #11, opened 2017-11-02
Add upper bound as input to treewidth branch and bound

aidanproy:
Also working on implementations of pruning strategies (Theorems 6.2,6.3...)


aidanproy:
These changes are ready to be pulled

PR #17, opened 2017-12-06
Min weighted vertex cover closes #6

### LINKED ISSUE #6: min_weighted_vertex_cover
dwave_networkx currently implements min_vertex_cover, it would be relatively straightforward to generalise it to min_weighted_vertex_cover as in networkx.

bellert:
Weighted version of `min_vertex_cover()`
- Since `min_vertex_cover()` was implemented by calling `maximum_independent_set()`, implemented `maximum_weighted_independent_set()` which is used by `min_weighted_vertex_cover()`
- Both `min_vertex_cover()` and `maximum_independent_set()` now are implemented by calling their weighted versions with `weight=None`

Closes #6

### DIFF
--- a/dwave_networkx/algorithms/independent_set.py
+++ b/dwave_networkx/algorithms/independent_set.py
@@ -1,6 +1,90 @@
 from dwave_networkx.utils import binary_quadratic_model_sampler
 
-__all__ = ["maximum_independent_set", "is_independent_set"]
+__all__ = ["maximum_weighted_independent_set", "maximum_independent_set", "is_independent_set"]
+
+
+@binary_quadratic_model_sampler(2)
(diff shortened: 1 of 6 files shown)


dwave_networkx/algorithms/independent_set.py
```diff
@@ -104,3 +151,65 @@ def is_independent_set(G, indep_nodes):
 
     """
     return not bool(G.subgraph(indep_nodes).edges)
+
+
+@binary_quadratic_model_sampler(2)
+def _weighted_independent_sets(G, weight=None, sampler=None, **sampler_args):
```
arcondello: I would change the name of this function to better identify it as internal.

PR #21, opened 2018-02-05
New examples

sploiber:


dwave_networkx/examples/min_vertex_color.py
```diff
@@ -0,0 +1,18 @@
...
+
+# Set up a Networkx Graph
+G = nx.Graph()
+G.add_edges_from([(0,1), (0,2), (1,2), (1,3), (2,3), (1,4), (2,4), (3,4), (3,5), (4,5), (5,2)])
+
+# Get the minimum vertex coloring, which is known in this case to be of
+# length 6
+candidate = dnx.min_vertex_coloring(G, sampler)
+if dnx.is_vertex_coloring(G, candidate) and len(candidate) == 6:
+  print str(candidate) + " is a minimum vertex coloring"
```
arcondello: To run on python 3, you'll need to use the new print format
`print(candidate, "is minimum vertex coloring")`

sploiber:
Alex, I made the changes, but in the output, in Python 2.7, it now looks
like this

([0, 3, 4, 5, 8], ' is a minimum vertex cover')

I'm OK with this - it gets the point across.
It works with Python 2.7 and Python 3.

But the overall question is - should the code be smarter than this, and
check whether the Python is 2.7 or 3?

Thanks-

Joel

arcondello:
Joel,

add `from __future__ import print_function` to the top of your file. I think that will fix it.

sploiber:
Alex, that did it beautifully. Thanks.

I've pushed it to sploiber/new_examples.

Does that properly update the code that has been "pull request"-ed?

Thanks-

Joel

dwave_networkx/examples/min_vertex_cover.py
```diff
@@ -0,0 +1,23 @@
...
+G = nx.Graph()
+G.add_edges_from([(0, 1), (0, 2), (1, 3), (2, 3), (3, 5), (4, 5), (3, 6), (4, 7), (5, 6), (5, 7), (6, 8), (7, 8), (8, 9)])
+
+# Get the minimum vertex cover, which is known in this case to be of
+# length 5
+candidate = dnx.min_vertex_cover(G, sampler)
+if dnx.is_vertex_cover(G, candidate) and len(candidate) == 5:
+  print (candidate, " is a minimum vertex cover")
+else: 
+  print (candidate, " is not a minimum vertex cover")
```
arcondello: It is a nitpick, but ideally no space between `print` and `(`.

Issue #31, opened 2018-02-18
Requirements Update


ccoffrin:
It seems like `draw_chimera` requires numpy and matplotlib, but these dependencies are not listed in `requirements.txt`.  That said, both are fairly heavy dependencies for a simply graph layout.  It might make sense for this package to provide the `chimera_layout` function, which then can be combined with any viz tool (e.g. matplotlib) to provide the drawing dependencies.

arcondello:
This is consistent with Networkx's approach which (for this package) we are trying to match as closely as possible. However when I tried to cite somewhere in Networkx's documentation that makes this clear I couldn't find anything. This is the closest (https://networkx.github.io/documentation/stable/reference/drawing.html) I'll make sure we update our documentation accordingly.

JoelPasvolsky:
Updated RtD. I'll assign back to you so you can add extra requirements on numpy & matplotlib. Here's what NetworkX has:
networkx/requirements/extras.txt
numpy>=1.12.0
scipy>=0.19.0
pandas>=0.20.1
matplotlib>=2.0.2
pygraphviz>=1.3.1

PR #39, opened 2018-02-20
Update setup.py extras to include matplotlib and numpy

arcondello:
* Also update travis to skip superfluous deploys


arcondello:
Addresses #31

PR #55, opened 2018-05-11
Add draw_chimera_embedding

aidanproy:
Adds a function that draws embeddings in chimera with matplotlib (similar to draw_chimera, but colors represent distinct chains instead of bqm model weights).


arcondello:
Closed by #56

PR #62, opened 2018-06-05
Add lagrange

vgoliber:
Added Lagrange parameter to independent_set.py as an optional parameter.


dwave_networkx/algorithms/independent_set.py
```diff
@@ -6,7 +6,7 @@
 
 
 @binary_quadratic_model_sampler(2)
-def maximum_weighted_independent_set(G, weight=None, sampler=None, **sampler_args):
+def maximum_weighted_independent_set(G, weight=None, sampler=None, Lagrange=2.0, **sampler_args):
```
arcondello: `lagrange` should be lower case https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

PR #66, opened 2018-07-18
Add TSP Function

vgoliber:
Created new function for traveling salesman problem


dwave_networkx/algorithms/TSP.py
```diff
@@ -0,0 +1,157 @@
...
+       List of nodes in order to be visited on a route
+
+    Notes
+    -----
+    Samplers by their nature may not return the optimal solution. This
+    function does not attempt to confirm the quality of the returned
+    sample.
+
+    References
+    ----------
```
arcondello: If there are no references this can be deleted

dwave_networkx/algorithms/TSP.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # nodes that are spin up or true are exactly the ones in S.
+    return route
+
+def traveling_salesman_qubo(G, lagrange=2.0):
+    """Return the QUBO with ground states corresponding to a minimum TSP route.
+    Parameters
+    ----------
+    G : NetworkX graph
+
```
arcondello: I think sphinx will choke if we have a line here

dwave_networkx/algorithms/TSP.py
```diff
@@ -0,0 +1,157 @@
...
+            j = node%N
+            v = (node-j)/N
+            route[j] = int(v)
+
+    # nodes that are spin up or true are exactly the ones in S.
+    return route
+
+def traveling_salesman_qubo(G, lagrange=2.0):
+    """Return the QUBO with ground states corresponding to a minimum TSP route.
+    Parameters
```
arcondello: Add whitespace here

dwave_networkx/algorithms/TSP.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def index(a, b):
+        return (a)*N+(b)
+
+    ## Creating the QUBO
+    # Start with an empty QUBO
+    Q = {}
+    for i in range(N*N):
+        for j in range(N*N):
+            Q.update({(i,j): 0})
```
arcondello: This block can be replaced with.

`Q = {(i, j): 0.0 for i in range(N*N) for j in range(N*N)}`

In general the `.update` method is slow and should be avoided for single key/value pair updates.

dwave_networkx/algorithms/TSP.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Constraint that each row has exactly one 1
+    for v in range(N):
+        for j in range(N):
+            Q[(index(v,j), index(v,j))] += -1*lagrange
```
arcondello: because lagrange is defaulted to a float, float multiplication in python in is faster. So for instance
`Q[(index(v,j), index(v,j))] += -lagrange` or even `Q[(index(v,j), index(v,j))] -= lagrange`
and
`Q[(index(v,j), index(v,k))] += 2.0*lagrange`

the performance gain is tiny so it doesn't matter here really, but it's good practice :slightly_smiling_face:

arcondello:
We also absolutely need unittests and this function needs to be added to the documentation

dwave_networkx/algorithms/TSP.py
```diff
@@ -0,0 +1,157 @@
...
+    # Objective that minimizes distance
+    for u in range(N):
+        for v in range(N):
+            if u!=v:
+                for j in range(N):
+                    Q[(index(u,j), index(v,(j+1)%N))] += G[u][v]['weight']
+
+    return Q
+
+def is_valid_route(G, route):
```
arcondello: perhaps `is_hamitonian_path` to make this less ambiguous.

arcondello:
looks good to me!

PR #76, opened 2018-10-11
Add canonical_chimera_labelling

arcondello:


dwave_networkx/algorithms/canonicalization.py
```diff
@@ -0,0 +1,161 @@
...
+        A mapping from the current labels to a 4-tuple of Chimera indices.
+
+    """
+    adj = G.adj
+
+    if t is None:
+        if hasattr(G, 'edges'):
+            num_edges = len(G.edges)
+        else:
+            num_edges = len(G.quadratic)
```
m3ller: Potential five lines to one:
```
num_edges = len(G.edges) if hasattr(G, 'edges') else len(G.quadratic)
```

dwave_networkx/algorithms/tests/test_canonicalization.py
```diff
@@ -0,0 +1,124 @@
...
+
+class TestCanonicalChimeraLabeling(unittest.TestCase):
+    def test_tile_identity(self):
+        C1 = dnx.chimera_graph(1)
+        coord = chimera_coordinates(1, 1, 4)
+
+        labels = canonical_chimera_labeling(C1)
+        labels = {v: coord.int(labels[v]) for v in labels}
+
+        G = nx.relabel_nodes(C1, labels, copy=True)
```
m3ller: Are we following math notation for variable naming? As in capital `G` for graphs and capital `C` for chimera graphs?

dwave_networkx/algorithms/canonicalization.py
arcondello: I think the 5 lines is clearer, but can be convinced otherwise.

dwave_networkx/algorithms/tests/test_canonicalization.py
arcondello: Following NetworkX's notation for this library. Also why we use a different docstring convention.

dwave_networkx/algorithms/canonicalization.py
randomir: Haha, if you want even shorter (and unclearer):
```
num_edges = len(getattr(G, 'edges', None) or G.quadratic))
```
(assuming G.edges is never 0)

PR #112, opened 2019-04-17
Draw qubit graph yield

joseppinilla:
Simple implementation to highlight faults in Chimera or Pegasus graphs.
Example usage:
```
G = dnx.chimera_graph(2, 2, 4, data=False)
G.remove_edges_from([(0,6),(10,13),(26,31)])
G.remove_nodes_from([18,23])
dnx.draw_chimera_yield(G)
plt.show()
```

```
G = dnx.pegasus_graph(3, data=False)
G.remove_edges_from([(5,104),(12,96),(23,112)])
G.remove_nodes_from([109,139])
dnx.draw_pegasus_yield(G)
plt.show()
```

It's also possible to use before a draw_..._embedding() because it differentiates faults with node shape ='x' and edge style = 'dashed'

Also modified description of unused_color which allows color strings and a minor typo in test_chimera_layout.

### DIFF
--- a/dwave_networkx/drawing/chimera_layout.py
+++ b/dwave_networkx/drawing/chimera_layout.py
@@ -310,7 +310,7 @@ def draw_chimera_yield(G, **kwargs):
 
     fault_shape : string, optional (default='x')
         The shape of the fault nodes. Specification is as matplotlib.scatter
-        marker, one of ‘so^>v<dph8’.
+        marker, one of 'so^>v<dph8'.
 
     fault_style : string, optional (default='dashed')
         Edge fault line style (solid|dashed|dotted,dashdot)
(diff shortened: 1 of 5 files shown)


dwave_networkx/drawing/chimera_layout.py
```diff
@@ -23,8 +23,8 @@
 from networkx import draw
```
arcondello: Looks like you need to add `# coding=utf-8` to the top of this file to fix the tests in python2. See https://stackoverflow.com/questions/10589620/syntaxerror-non-ascii-character-xa3-in-file-when-function-returns-%C2%A3

joseppinilla: Is this something you'll be needing in every file?
I'm guessing this is because of the description of matplotlib's scatter.
Is it enough to just remove the characters that cause this?

arcondello: Yes, removing the characters would do it. It looks like the problem is actually in `dwave_networkx/drawing/qubit_layout.py`

joseppinilla: OK, I'll make the changes and make sure it passes. I don't want to add the encoding line if it's not necessary.

PR #114, opened 2019-04-22
Make relabelling methods more explicit

### LINKED ISSUE #101: doc issue: tuple_pairs copy pasted from int_pairs
minor doc bug. 
tuple_pairs docs is copy pasted from int_pairs. 

https://github.com/dwavesystems/dwave_networkx/blob/197539faf609f9bfd22d2b62ea5b112277327a47/dwave_networkx/generators/chimera.py#L482-L496

### LINKED ISSUE #113: Create a more generic coordinates object for chimera and pegasus
There are now three different labeling systems for Pegasus
* linear: see https://github.com/dwavesystems/dwave_networkx/blob/dae190a5530d9f06fca89bf7e3737c06c7e6ab5a/dwave_networkx/generators/pegasus.py#L460
* pegasus: see https://github.com/dwavesystems/dwave_networkx/blob/dae190a5530d9f06fca89bf7e3737c06c7e6ab5a/dwave_networkx/generators/pegasus.py#L479
* 'nice' as introduced in #95 and revised in #106 

So we need a many-to-many translator.

One approach suggested by @boothby would be something like (using grids as an example):
```
>>> f = grid_coordinates.from_xy(shape=(10, 10)).to_linear()
>>> f(0, 0)
0
```

This should also support chimera and pegasus. So something like
```
>>> f = pegasus_coordinates.from_chimera(shape=(16, 16, 4)).to_nice(t=1)
>>> f(0, 0, 0, 0)
(1, 0, 0, 0, 0)
```

arcondello:
Closes #101 
Closes #113 though not with full generality

### DIFF
--- a/dwave_networkx/generators/chimera.py
+++ b/dwave_networkx/generators/chimera.py
@@ -26,7 +26,8 @@ from networkx import diameter
 from dwave_networkx import _PY2
 from dwave_networkx.exceptions import DWaveNetworkXException
 
-__all__ = ['chimera_graph', 'find_chimera_indices', 'chimera_to_linear',
+__all__ = ['chimera_graph', 'chimera_coordinates', 'find_chimera_indices',
+           'chimera_elimination_order', 'chimera_to_linear',
            'linear_to_chimera']
 
 # compatibility for python 2/3
@@ -357,6 +358,25 @@ class chimera_coordinates(object):
         t : int, optional (default 4)
             The size of the shore within each Chimera tile.
 
+        Examples
+        --------
+
+        Convert between Chimera coordinates and linear indices directly
+
+        >>> coords = dnx.chimera_coordinates(16, 16, 4)
+        >>> coords.chimera_to_linear((0, 2, 0, 1))
+        17
+        >>> coords.linear_to_chimera(17)
+        (0, 2, 0, 1)
+
+        Construct a new graph with the coordinate labels
+
+        >>> C16 = dnx.chimera_graph(16)
+        >>> coords = dnx.chimera_coordinates(16)
+        >>> G = nx.Graph()
+        >>> G.add_nodes_from(coords.iter_linear_to_chimera(C16.nodes))
+        >>> G.add_edges_from(coords.iter_linear_to_chimera_pairs(C16.edges))
+
         See also
         --------
         :func:`.chimera_graph` for a description of the various conventions.
@@ -403,8 +423,8 @@ class chimera_coordinates(object):
         return self.iter_chimera_to_linear(qlist)
 
     def iter_chimera_to_linear(self, qlist):
-        """Return an iterator converting 4-term Chimera coordinates to linear
-        indices.
+        """Return an iterator converting a sequence of 4-term Chimera
+        coordinates to linear indices.
         """
         m, n, t = self.args
         for (i, j, u, k) in qlist:
@@ -419,8 +439,8 @@ class chimera_coordinates(object):
         return self.iter_linear_to_chimera(rlist)
 
     def iter_linear_to_chimera(self, rlist):
-        """Return an iterator converting linear indices to 4-term Chimera
-        coordinates.
+        """Return an iterator converting a sequence of linear indices to 4-term
+        Chimera coordinates.
         """
         m, n, t = self.args
         for r in rlist:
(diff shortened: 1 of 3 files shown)


arcondello:
Closes #113

dwave_networkx/generators/chimera.py
```diff
@@ -356,147 +357,119 @@ def __init__(self, m, n=None, t=4):
...
     def ints(self, qlist):
+        """Deprecated alias for `iter_chimera_to_linear`."""
+        msg = ('chimera_coordinates.ints is deprecated and will be removed in '
+               'dwave-networkx 0.9.0, please use '
+               'chimera_coordinates.iter_chimera_to_linear instead')
+        warnings.warn(msg, DeprecationWarning)
+        return self.iter_chimera_to_linear(qlist)
+
+    def iter_chimera_to_linear(self, qlist):
+        """Return an iterator converting 4-term Chimera coordinates to linear
```
JoelPasvolsky: I think this needs to specify that the input is a list, something like, "converting a list of 4-term Chimera coordinates"

dwave_networkx/generators/chimera.py
```diff
@@ -356,147 +357,119 @@ def __init__(self, m, n=None, t=4):
...
+        """Deprecated alias for `iter_linear_to_chimera`."""
+        msg = ('chimera_coordinates.tuples is deprecated and will be removed in '
+               'dwave-networkx 0.9.0, please use '
+               'chimera_coordinates.iter_linear_to_chimera instead')
+        warnings.warn(msg, DeprecationWarning)
+        return self.iter_linear_to_chimera(rlist)
+
+    def iter_linear_to_chimera(self, rlist):
+        """Return an iterator converting linear indices to 4-term Chimera
+        coordinates.
```
JoelPasvolsky: I think this needs to specify that the input is a list, something like, "converting a list of linear"

dwave_networkx/generators/chimera.py
```diff
@@ -356,147 +357,119 @@ def __init__(self, m, n=None, t=4):
...
+        """Deprecated alias for `iter_chimera_to_linear_pairs`."""
+        msg = ('chimera_coordinates.int_pairs is deprecated and will be removed in '
+               'dwave-networkx 0.9.0, please use '
+               'chimera_coordinates.iter_chimera_to_linear_pairs instead')
+        warnings.warn(msg, DeprecationWarning)
+        return self.iter_chimera_to_linear_pairs(plist)
+
+    def iter_chimera_to_linear_pairs(self, plist):
+        """Return an iterator converting pairs of 4-term Chimera coordinates to
+        pairs of linear indices.
```
JoelPasvolsky: At this point I worry users might find getting the format correct hard without it being specified because it is not intuitive that plist is a list that contains one or more tuples in which each member is a 4-term tuple. If we want to keep it short maybe either a 2nd sentence describing the format or a one-line example?

dwave_networkx/generators/chimera.py
```diff
@@ -356,147 +357,119 @@ def __init__(self, m, n=None, t=4):
...
-    def tuple_pairs(self, plist):
+        msg = ('chimera_coordinates.tuple_pairs is deprecated and will be removed in '
+               'dwave-networkx 0.9.0, please use '
+               'chimera_coordinates.iter_linear_to_chimera_pairs instead')
+        warnings.warn(msg, DeprecationWarning)
+        return self.iter_linear_to_chimera_pairs(plist)
+
+    def iter_linear_to_chimera_pairs(self, plist):
+        """Return an iterator converting pairs of linear indices to pairs of
+        4-term Chimera coordinates.
```
JoelPasvolsky: Same comment as for iter_chimera_to_linear_pairs()

dwave_networkx/generators/chimera.py
```diff
@@ -356,147 +357,119 @@ def __init__(self, m, n=None, t=4):
             The number of columns in the Chimera lattice.
         t : int, optional (default 4)
```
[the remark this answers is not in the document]
arcondello: Because if None is provided it is set to 4

dwave_networkx/generators/chimera.py
arcondello: You're right, in the pegasus ones it says "...converting a sequence of 4-term...", will fix.

dwave_networkx/generators/chimera.py
arcondello: I added an example showing the use

joseppinilla:
I'm curious, both chimera and pegasus will still have `labels` options `'coordinate'` and `'int'`.
Is there a reason `'int'` shouldn't be replaced with `'linear'` to match the new explicit relabelling methods?

PR #116, opened 2019-04-29
Add colorbar toggling based on biases provided

conta877:


dwave_networkx/drawing/qubit_layout.py
```diff
@@ -133,20 +133,31 @@ def node_color(v):
         if edge_vmax is None:
             edge_vmax = vmag
 
-    draw(G, layout, nodelist=nodelist, edgelist=edgelist,
+    fig = plt.figure()
```
arcondello: I think we still want to do `figure(1)` here

conta877:
@arcondello what do you think of adding ax/cax kwargs? (pushed)

PR #130, opened 2019-08-16
Functionality for converting between linear and chimera indexing.

davage:
Added functions for linear_to_chimera and chimera_to_linear index conversion, as well as unit tests.

### DIFF
--- a/dwave_networkx/generators/chimera.py
+++ b/dwave_networkx/generators/chimera.py
@@ -24,7 +24,9 @@ from networkx import diameter
 from dwave_networkx import _PY2
 from dwave_networkx.exceptions import DWaveNetworkXException
 
-__all__ = ['chimera_graph', 'find_chimera_indices', 'chimera_elimination_order', 'chimera_to_linear', 'linear_to_chimera']
+__all__ = ['chimera_graph', 'find_chimera_indices',
+           'chimera_elimination_order', 'chimera_to_linear',
+           'linear_to_chimera']
 
 # compatibility for python 2/3
 if _PY2:
@@ -518,7 +520,8 @@ def linear_to_chimera(r, m, n=None, t=None):
     j : int
         The row of the Chimera index's unit cell associated with `r`.
     u : int
-        Whether the index is even (0) or odd (1); the side of the bi-partite graph of the Chimera unit cell.
+        Whether the index is even (0) or odd (1); the side of the bi-partite
+        graph of the Chimera unit cell.
     k : int
         Index into the Chimera unit cell.
 
@@ -541,7 +544,7 @@ def linear_to_chimera(r, m, n=None, t=None):
     return i, j, u, k
 
 
-def chimera_to_linear(i, j, u, k, m, n, t):
+def chimera_to_linear(i, j, u, k, m, n=None, t=None):
     """Convert the chimera index `(i, j, u, k)` into a linear index.
 
     Parameters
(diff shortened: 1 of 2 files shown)


dwave_networkx/generators/chimera.py
```diff
@@ -495,3 +495,83 @@ def tuple_pairs(self, plist):
             Equivalent to (tuple(self.tuples(p)) for p in plist)
         """
         return self.__pair_repack(self.tuples, plist)
+
+def linear_to_chimera(r, m, n=None, t=None):
+    """Convert the linear index `r` into a chimera index.
+
+
```
arcondello: Should only be one space here

dwave_networkx/generators/chimera.py
```diff
@@ -495,3 +495,83 @@ def tuple_pairs(self, plist):
...
+    j : int
+        The row of the Chimera index's unit cell associated with `r`.
+    u : int
+        Whether the index is even (0) or odd (1); the side of the bi-partite graph of the Chimera unit cell.
+    k : int
+        Index into the Chimera unit cell.
+
+    Examples
+    --------
+    >>> G = dnx.linear_to_chimera(0, 8, 8, 4)
```
arcondello: Should be a space between the `------` and the `>>> ...`, also the example should show the output. E.g.
```
Examples
--------

>>> G = dnx.linear_to_chimera(0, 8, 8, 4)
(0, 0, 0, 0)
```

dwave_networkx/generators/chimera.py
davage: One empty line here right?

dwave_networkx/generators/chimera.py
davage: Do we want to make the spacing congruent with this in the Returns section, or just put the blank line here?

dwave_networkx/generators/chimera.py
arcondello: Yes

dwave_networkx/generators/chimera.py
arcondello: I think it only needs to be there for examples, but consistency is good!

dwave_networkx/generators/chimera.py
```diff
@@ -495,3 +495,83 @@ def tuple_pairs(self, plist):
...
+    if t is None:
+        t = 4
+
+    r, k = divmod(r, t)
+    r, u = divmod(r, 2)
+    i, j = divmod(r, n)
+    return i, j, u, k
+
+
+def chimera_to_linear(i, j, u, k, m, n, t):
```
arcondello: I think it would be good to have the `n` and `t` inputs consistent with `lienar_to_chimera`, i.e. `n=m, t=4` by default.

dwave_networkx/generators/chimera.py
```diff
@@ -24,7 +24,7 @@
 from dwave_networkx import _PY2
 from dwave_networkx.exceptions import DWaveNetworkXException
 
-__all__ = ['chimera_graph', 'find_chimera_indices', 'chimera_elimination_order']
+__all__ = ['chimera_graph', 'find_chimera_indices', 'chimera_elimination_order', 'chimera_to_linear', 'linear_to_chimera']
```
arcondello: 80 char width

dwave_networkx/generators/chimera.py
```diff
@@ -495,3 +495,83 @@ def tuple_pairs(self, plist):
...
+
+
+    Returns
+    -------
+    i : int
+        The column of the Chimera index's unit cell associated with `r`.
+    j : int
+        The row of the Chimera index's unit cell associated with `r`.
+    u : int
+        Whether the index is even (0) or odd (1); the side of the bi-partite graph of the Chimera unit cell.
```
arcondello: 80 char width

arcondello:
See also #113

PR #131, opened 2019-08-28
Error in formula

schnorea:
The comment reads like this:
`pegasus_graph(m, fabric_only=True) contains :math:`24m(m-1)-8m` nodes, and`

should this not read:
`pegasus_graph(m, fabric_only=True) contains :math:`24m(m-1)-8(m-1)` nodes, and`

I put 16 into the formula and get 5632 I put 16 into the function and get 5640.


arcondello:
Note that the CI failures are not caused by this PR

Issue #12, opened 2017-11-03
test failure for min_maximal_matching_typical test


arcondello:
======================================================================
FAIL: test_min_maximal_matching_typical (dwave_networkx.algorithms.tests.test_matching.TestMatching)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/dwavesystems/dwave_networkx/dwave_networkx/algorithms/tests/test_matching.py", line 435, in test_min_maximal_matching_typical
    "nodes: {}\nedges:{}".format(G.nodes(), G.edges()))
AssertionError: False is not true : nodes: [0, 1, 2, 3, 4, 5, 6]
edges:[(1, 6), (2, 5), (3, 4)]
----------------------------------------------------------------------
Ran 76 tests in 22.659s
FAILED (failures=1)

arcondello:
Test failure appeared in travis build for open pull request that did not alter the min_maximal_matching code

arcondello:
This exception happened again in another unrelated PR (#133) https://circleci.com/gh/dwavesystems/dwave_networkx/1603

PR #133, opened 2019-09-09
Elimination order refresh

### LINKED ISSUE #28: algorithms/elimination's max_cardinality_heuristic() and elimination_order_width()
* Parameter "inplace" in function max_cardinality_heuristic() seems to be just sitting there
* Functions max_cardinality_heuristic() and elimination_order_width() should either be added to the TOC in elimination.rst to expose them or a comment should be added explaining why not

### LINKED ISSUE #29: algorithms/elimination's treewidth_branch_and_bound() assert
The docstring says "treewidth_upperbound : int ...using this parameter can result in no returned order."

But actually using it with a value smaller than the real upperbound causes it to assert:

In [265]: dnx.treewidth_branch_and_bound(K_7,[0,1,2,3,4,5,6],6)
Out[265]: (6, [0, 1, 2, 3, 4, 5, 6])

In [266]: dnx.treewidth_branch_and_bound(K_7,[0,1,2,3,4,5,6],5)

---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-266-9e0ef4d24603> in <module>()
----> 1 dnx.treewidth_branch_and_bound(K_7,[0,1,2,3,4,5,6],5)

/usr/local/lib/python2.7/dist-packages/dwave_networkx/algorithms/elimination_ordering.pyc in treewidth_branch_and_bound(G, elimination_order, treewidth_upperbound)
    429     # if our upper bound is the same as f, then we are done! Otherwise begin the
    430     # algorithm
--> 431     assert f <= ub, "Logic error"
    432     if f < ub:
    433         # we need only deal with the adjacency structure of G. We will also

### LINKED ISSUE #48: elimination orderings don't work properly on graphs with self loops
Networkx graphs are allowed to have loops by default. dwnx.elimination_ordering.min_width_heuristic and min_fill_heuristic (and maybe others) don't work properly in those cases, so this condition should be checked before starting the algorithm.

### LINKED ISSUE #115: Move chimera_elimination_order from generators to algos/elimination_order
**Description**
currently it's here: https://docs.ocean.dwavesys.com/projects/dwave-networkx/en/latest/reference/generated/dwave_networkx.generators.chimera_elimination_order.html#dwave_networkx.generators.chimera_elimination_order

It should be here: https://docs.ocean.dwavesys.com/projects/dwave-networkx/en/latest/reference/algorithms/elimination_ordering.html

**To Reproduce**
open the github/RTD

**Expected behavior**
Easily found

**Environment:**
 - OS: all
 - Python version: all

**Additional context**
None

arcondello:
Closes #28 
Closes #29 
Closes #48 
Closes #115


dwave_networkx/algorithms/elimination_ordering.py
```diff
@@ -125,7 +134,7 @@ def minor_min_width(G):
     """
     # we need only deal with the adjacency structure of G. We will also
     # be manipulating it directly so let's go ahead and make a new one
-    adj = {v: set(G[v]) for v in G}
+    adj = {v: set(u for u in G[v] if u != v) for v in G}
```
JoelPasvolsky: Just wondering, seeing as this is done 6 times in this module, why not define it as an internal func?
```
def _get_adj(G):
     """<condensed version of> #we need only deal with the adjacency structure of G. We will also
         # be manipulating it directly so let's go ahead and make a new one
    """
    return {v: set(u for u in G[v] if u != v) for v in G}
```

dwave_networkx/algorithms/elimination_ordering.py
```diff
@@ -185,8 +188,6 @@ def min_fill_heuristic(G):
     This example computes an upper bound for the treewidth of the :math:`K_4`
     complete graph.
 
-    >>> import dwave_networkx as dnx
-    >>> import networkx as nx
```
JoelPasvolsky: Will users know what "nx" and "dnx" represent without the import? Shouldn't the following lines be updated to something like:
```
 >>> K_4 = networkx.complete_graph(4)
 >>> tw, order = dwave_networkx.min_fill_heuristic(K_4)
```
Same for following examples

arcondello:
Failure appears to be due to #12 and not this PR

dwave_networkx/algorithms/elimination_ordering.py
arcondello: I think in this library it is safe to assume that they know. NetworkX follows this convention (to pick a random example https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.bridges.bridges.html#networkx.algorithms.bridges.bridges)

dwave_networkx/algorithms/elimination_ordering.py
arcondello: In this case I think one line functions are better duplicated, especially when they are not doing complicated logic. Especially since we would probably need to keep the comment string in place - I cannot think of a nice way to name that's more explicit than the line itself.

PR #135, opened 2019-11-08
Use gcf and gca to get/create figure axes in qubit graph

joseppinilla:
I noticed that using `dnx.draw_chimera` under a `plt.subplot(...)` would fail because of the use of `plt.figure(1)` inside `draw_qubit_graph`.

I think it can be fixed this way.
plt.gcf(): _Get the current figure. If no current figure exists, a new one is created using figure()._

Then, plt.gca() is also necessary. The `cax` is OK as is.


conta877:
looks good!

PR #137, opened 2019-11-20
change the way colormaps created for graph drawing

conta877:
instead of using only the minimum and maximum values of a given dataset (linear or quadratic biases in this case), if midpoint is provided the colormap will have its middle value at the midpoint


arcondello:
Would it maybe make more sense to just calculate the midpoint based on the absolute value of the min/max?

conta877:
Thats what the function does normally without the midpoint input. but for data visualization (instead of bias provided), one may want the mid point to correspond to data mean or median

PR #138, opened 2019-11-22
Adding Lagrange parameter to min_vertex_cover and lower-level covers

sploiber:


dwave_networkx/algorithms/cover.py
```diff
@@ -125,7 +131,7 @@ def min_vertex_cover(G, sampler=None, **sampler_args):
     >>> sampler = dimod.ExactSolver()  # small testing sampler
     >>> G = dnx.chimera_graph(1, 1, 4)
     >>> G.remove_node(7)  # to give a unique solution
-    >>> dnx.min_vertex_cover(G, sampler)
+    >>> dnx.min_vertex_cover(G, lagrange, sampler)
```
arcondello: Think you want `lagrange=2.0` or just drop that parameter from the example entirely.

sploiber: I did want lagrange=2.0, or the param can be ignored. I'm OK with both.

arcondello: Either works for me

PR #143, opened 2019-12-05
removed calls to deprecated coordinate translators

boothby:
I got sick of seeing deprecation warnings, so I cleaned out internal functions to avoid them.


arcondello:
That's my bad, should have been part of #114 . Thank you!

PR #146, opened 2019-12-06
Fix error in QUBO constraint

### LINKED ISSUE #145: TSP QUBO Error
Lines 170-171 in tsp.py adds QUBO coefficient for this specific constraint both above *and* below the diagonal in the Q matrix.

Code needs to be corrected to put coefficient in one place or the other (but not both).

vgoliber:
Constraint on lines 170-171 had quadratic coefficient placed both above and below the diagonal.  Corrected error by dividing coefficient by 2.  Added comments for description.

Resolves #145

### DIFF
--- a/dwave_networkx/algorithms/tests/test_tsp.py
+++ b/dwave_networkx/algorithms/tests/test_tsp.py
@@ -215,8 +215,7 @@ class TestTSPQUBO(unittest.TestCase):
     
     def test_weighted_complete_graph(self):
         G = nx.Graph()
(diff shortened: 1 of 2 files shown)


arcondello:
It would be good if this came with a test that failed in the old version and now passes. Do we have one we can add?

vgoliber:
@arcondello new test added to check total sum of QUBO entries is what we expect on a smaller graph.

arcondello:
Looks good to me! Thanks

Issue #147, opened 2019-12-09
TSP route check incorrect


vgoliber:
In tsp.py, a check for a valid route returns `true` if and only if `(len(route) == len(set(G)))` is true (line 214).  However, since route is initialized (line 104) as `route = [None]*len(G)`, the statement will always be true.  

The feasibility check is always returning true, regardless of whether or not the returned answer is a valid route.

Issue #103, opened 2019-04-01
TSP - why inconsistent?


neverwiredhouse:
I am curious what causes it to be wrong almost always by the quantum computer? It seems like it's not ready yet, but I am more just curious the difficulties presented by this problem.

sploiber:
Hello @neverwiredhouse, could you please share an example in which you found a wrong answer?

neverwiredhouse:
@sploiber ,
No problem. Thanks for the fast reply.
[here is a gist](https://gist.github.com/neverwiredhouse/3cb8fe97de856bc43bcaca63da4bf5e4)

pretty simple. I try using the "exact" method, which is on the PC, and then the quantum method and the answer is always random.

sploiber:
Hello @neverwiredhouse, yes indeed, I am also seeing strange behavior from the QPU example. I will write it up below, but I will also document the overall picture thoroughly, to be sure we're on the same page. 
First, here's the exact solver code:
```
import dwave_networkx as dnx
import networkx as nx
from dimod import ExactSolver
G = nx.complete_graph(4)
G.add_weighted_edges_from({(0, 1, 1), (0, 2, 2), (0, 3, 3), (1, 2, 3), (1, 3, 4), (2, 3, 5)})
sampler = ExactSolver()
answer = dnx.traveling_salesman(G, sampler)
print(answer)
```
Running in Python 3.7, on a Mac with Anaconda, I get an answer [2, 1, 0, 3]. If I add up all the paths in the graph (2->1, 1->0, 0->3, and 3->2), I get a distance of 12. This does seem to be the minimum answer, so ExactSolver seems to be working.

If I run with SimulatedAnnealingSampler instead, I get different paths sometimes, but the distance is always 12. This is consistent - the different solver is finding a valid answer - just a different path.

Now, if I look at the code with the QPU:
```
import dwave_networkx as dnx
import networkx as nx
from dwave.system.samplers import DWaveSampler
from dwave.system.composites import EmbeddingComposite
G = nx.complete_graph(4)
G.add_weighted_edges_from({(0, 1, 1), (0, 2, 2), (0, 3, 3), (1, 2, 3), (1, 3, 4), (2, 3, 5)})
sampler = EmbeddingComposite(DWaveSampler())
answer = dnx.traveling_salesman(G, sampler, 1)
print(answer)
```
I do indeed get unpredictable results - sometimes long vectors - sometimes vectors of length one. I'll continue on this today.

[...]

Issue #82, opened 2018-12-13
TSP needs more usage info


JoelPasvolsky:
**Description**
Most TSP problems I try fail to produce a minimal route. The parameters and working range of splvable problems should be outlined. 

**To Reproduce**
The following alternative weights to the 5-node problem, EdgeView([(0, 1), (0, 2), (0, 3), (1, 2), (1, 3), (2, 3)]), all fail:

* [{'weight': 22.36},  {'weight': 10.0},  {'weight': 14.14},  {'weight': 20.0},  {'weight': 22.36}, {'weight': 22.36}]
  produces route: [3, 1, 0] (non-Hamiltonian route)
* [{'weight': 2.24}, {'weight': 1.0}, {'weight': 1.41}, {'weight': 2.0}, {'weight': 2.24}, {'weight': 2.24}]
  produces route: [1, 0, 3, 2] (a better answer is [1 2 0 3])
* [{'weight': 2}, {'weight': 4}, {'weight': 1}, {'weight': 2}, {'weight': 4}, {'weight': 3}]
  produces route: [3, 2, 1, 0] (a better answer is [2 1 0 3])

**Expected behavior**
More info on range of problem inputs, how to set lagrange, etc

**Environment:**
 - OS: Windows
 - Python version: 2.7

**Additional context**
@vgoliber

vgoliber:
Choosing a value for a Lagrange parameter is a topic that is relevant for all of our algorithm tools (not just TSP).  I think it would be more appropriate to provide higher level guidance and/or references on this subject as a whole rather than guidance specific to each algorithm.

JoelPasvolsky:
Agree, @vgoliber, you mentioned having references and some guidance. Please add that here for that general update. Thanks!

PR #148, opened 2019-12-09
Update and improve TSP code

vgoliber:
Resolves issue #147 and also addresses issue #103.

There may be a faster way to check route validity, but this will get the job done.

### DIFF
--- a/dwave_networkx/algorithms/tsp.py
+++ b/dwave_networkx/algorithms/tsp.py
@@ -211,4 +211,4 @@ def is_hamiltonian_path(G, route):
 
     """
 
-    return (len(route) == len(set(G)))
+    return (set(route) == set(G))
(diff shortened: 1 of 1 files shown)


vgoliber:
Lagrange parameter update and validity check addresses issue #82 by providing a better default value and returning valid routes.

dwave_networkx/algorithms/tsp.py
```diff
@@ -148,6 +148,13 @@ def traveling_salesperson_qubo(G, lagrange=2, weight='weight'):
     """
     N = G.number_of_nodes()
 
+    if lagrange is None:
+        # if no lagrange parameter provided, set to 'average' tour length
+        if G.number_of_edges()>0:
+            lagrange = G.size(weight=weight)*G.number_of_nodes()/G.number_of_edges()
```
JoelPasvolsky: Could you add few words in the comment about why it makes sense to set this value for the parameter?

vgoliber: Added.

PR #151, opened 2019-12-23
Unified docs: add SDK index and move _static images

JoelPasvolsky:
Enables us to create a symlink to the docs folder in SDK and bring in selected documents without duplicating doc files in two index files. Also separates html and image folders.


JoelPasvolsky:
Preview in my RTD: https://dnx.readthedocs.io/en/latest/index.html

PR #158, opened 2020-02-27
Embedding drawing node size

joseppinilla:
The overlapped_embedding option is hogging the node_size parameter.
This PR allows the user to control it again, if wanted.


arcondello:
@joseppinilla , do you need a new dnx deploy or are you ok installing from master for now?

joseppinilla:
Installing from master should be OK for now.
Thanks!

PR #174, opened 2020-08-25
Standardizing package requirements files

hhtong:


tests/README.rst
```diff
@@ -0,0 +1 @@
+Tests can be found in the subdirectories of dwave_networkx/dwave_networkx.
```
hhtong, line 1: I'm not sure if this should be a readme or a comment in the requirements file.., or if there's a better way to do this.

randomir: This is fine, IMO.

randomir (approved):

PR #175, opened 2020-10-01
Add Exact Cover

hhtong:
Additional work that could be done:
- adding the graph representation of the problem (see [here](https://en.wikipedia.org/wiki/Exact_cover#Graph_representation))
- finding the smallest exact cover (least number of subsets)

### DIFF
--- a/dwave_networkx/algorithms/exact_cover.py
+++ b/dwave_networkx/algorithms/exact_cover.py
@@ -1,0 +1,174 @@
+# Copyright 2020 D-Wave Systems Inc.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+# ================================================================================================
+from dwave_networkx.utils import binary_quadratic_model_sampler
+from dimod import BinaryQuadraticModel
+
+__all__ = ['is_exact_cover', 'exact_cover']
+
+def is_exact_cover(problem_set, subsets):
+    """Determines whether the given list of subsets is an exact cover of
+    `problem_set`.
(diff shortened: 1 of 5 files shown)


dwave_networkx/algorithms/exact_cover.py
```diff
@@ -0,0 +1,174 @@
...
+                for j in range(i-1, -1, -1):
+                    if element in subsets[j]:
+                        bqm.add_interaction(i, j, 2)
+
+    bqm.offset = offset
+    
+    response = sampler.sample(bqm)
+
+    # Get lowest energy sample
+    sample = next(iter(response))
```
arcondello, line 163: this is an old syntax. You can just do `response.first.sample`

dwave_networkx/algorithms/exact_cover.py
```diff
@@ -0,0 +1,174 @@
...
+
+    result = []
+    for subset_index in sample:
+        if sample[subset_index] > 0:
+            result.append(subsets[subset_index])
+
+    # Validate that result is actually an exact cover
```
arcondello, line 170: Most of the other dnx functions return the solution whether it is correct or not, but document it.

dwave_networkx/algorithms/exact_cover.py
```diff
@@ -0,0 +1,174 @@
...
+    >>> dnx.is_exact_cover(problem_set, subsets)
+    False
+
+    """
+    # Verify that problem_set is valid
+    orig_len = len(problem_set)
+    problem_set = set(problem_set)
+    if len(problem_set) != orig_len:
+        raise ValueError("problem_set should not contain any duplicates")
+
```
randomir, line 62: Should you check the subsets are valid as well?

dwave_networkx/algorithms/exact_cover.py
```diff
@@ -0,0 +1,174 @@
...
+    if len(problem_set) != orig_len:
+        raise ValueError("problem_set should not contain any duplicates")
+    
+    # Verify that every element in problem_set is contained exactly once in subsets
+    cover = [x for subset in subsets for x in subset]
+
+    if len(cover) != orig_len:
+        return False
+
+    return all(x in cover for x in problem_set)
```
randomir, line 69: Shorter, and maybe a bit faster:
```suggestion
    return problem_set == set.union(*subsets) and orig_len == sum(map(len, subsets))
```

randomir (approved):
LGTM! Just a few minor suggestions.

PR #178, opened 2020-12-02
rm unnecesary future import

pau557:
I did this as part of a tutorial


dwave_networkx/examples/min_maximal_matching.py
```diff
@@ -13,7 +13,6 @@
 #    limitations under the License.
 #
 # ================================================================================================
-from __future__ import print_function
```
arcondello, line 16: Unfortunately we haven't disabled the CI for python 2.7, so this is failing.

pau557: sounds good

PR #182, opened 2020-12-08
Restructure tests

arcondello:
NetworkX keeps its tests in the same directory as the tested code. But putting them in root/tests/ is consistent with the rest of Ocean.


hhtong (approved):
/tests/README.rst can be removed, but otherwise LGTM!

PR #183, opened 2020-12-08
Clean out py2 boilerplate

arcondello:


hhtong (approved):

dwave_networkx/__init__.py
```diff
@@ -33,4 +27,5 @@
 import dwave_networkx.drawing
 from dwave_networkx.drawing import *
 
-from dwave_networkx.package_info import __version__, __author__, __authoremail__, __description__
+from dwave_networkx.package_info import __version__, __author__, \
```
randomir, line 30: When import spans multiple lines, I believe it's recommended to wrap it with parens.

arcondello: Eh... https://www.python.org/dev/peps/pep-0328/#id7 implies that parens are preferred. But not sure I take it as a stated recommendation

randomir: When they say:

> This part of the proposal had **BDFL approval** from the beginning.

and

> [alternatives are] **unpalatable** ...

that sounds like a recommendation to me :stuck_out_tongue:

arcondello: What's missing for me is the _why_ they are unpalatable. IMO the syntax above is cleaner. If it was in pep8 which is explicitly not especially motivated by anything except consistency, I would find it more compelling. Anyway, happy enough to use parens, though don't think I'll bother making another PR now.

randomir: No, I never meant to say you should change it to parens, just point out for future they are an option. A more robust option, IMO (one extra newline will break your import).

boothby: I can provide some opinions as an illustration of why the alternatives are unpalatable
1) backslash line-continuations to "avoid" long lines are gross because they're slightly fragile and fundamentally don't break the line, and I generally dislike the syntax
2) repeating the `from bla import blahblah` on multiple lines is because the repetition bugs me

PR #185, opened 2020-12-09
Convert to namespace package

arcondello:
Makes use of the new `dwave.plugins` namespace introduced with the release of [dwave-qiskit-plugin](https://github.com/dwavesystems/dwave-qiskit-plugin).

The old namespaces still work, but now raise a deprecation warning.

Todo: figure out what to do with the tests (which still import from the old namespace), more testing


arcondello:
For tests, there are three obvious options:
1) Continue testing using the old namespace under the assumption that it covers both
2) Switch to testing the new namespace and just test that the wrapper works
3) Test both, probably with parameterized or similar

I am leaning towards (2). Thoughts?

randomir (approved):
LGTM.

randomir:
@arcondello, I like option (3). Seems simplest (with parameterized) / no new tests needed, covers everything, and can be easily removed when we kill the old ns.

arcondello:
I am not sure I would describe it as easily removed, since you end up doing something like

```
import dwave.plugins.networkx as dnx
import dwave_networkx as old_dnx

@parameterized_class('dnx', [dnx, old_dnx])
class TestThingy(unittest.TestCase):
    def test_a(self):
        G = self.dnx.chimera_graph(4)

```

so removing means changing a lot of `self.dnx` to `dnx`. Not terrible but... I'll see if I can think of a nicer way.

randomir:
Also, we have quite a few tests, test files, and import statements like
```
from dwave_networkx.algorithms.canonicalization import rooted_tile
```
not just a simple `import dwave.plugins.networkx as dnx`.

And given python's import caching (and our namespace clone magic), I'm not 100% sure `dnx` and `old_dnx` wouldn't point to the same object.

Given all that, (2) might actually be easier.

arcondello:
Yeah, I am actually already halfway through refactoring a lot of that as a separate PR, because we're not really consistent. Worth doing in either case.

arcondello:
Pr mentioned above: https://github.com/dwavesystems/dwave-networkx/pull/186

PR #196, opened 2021-05-12
Update cookie methods  in conf.py

JoelPasvolsky:
`app.add_stylesheet()` and `app.add_javascript()` were renamed in version 1.8 and obsoleted in version 4

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,8 @@ jobs:
           command: |
             python -m venv env
             . env/bin/activate
-            pip install -U pip setuptools wheel twine
+            pip install -U pip setuptools wheel
+            pip install -U twine
             pip install -r requirements.txt -r tests/requirements.txt
 
       - save_cache: &save-cache-env
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -124,8 +124,8 @@ html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 html_static_path = ['_static']
 
 def setup(app):
-    app.add_stylesheet('cookie_notice.css')
-    app.add_javascript('cookie_notice.js')
+    app.add_css_file('cookie_notice.css')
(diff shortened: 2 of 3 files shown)


JoelPasvolsky:
Failure of `test-linux-3.6` unrelated to this PR

randomir:
Haha, `test-linux-3.6` is failing for a funny reason. Or, should I say for a confluence of reasons:

1) dependencies are not cached, so full install is attempted

2) one of dependencies, `cryptography` dropped `manylinux1` wheels (since our last successful build around Dec 2020), so pip 19+ is required to install from a `manylinux2010` wheel

3) the line to upgrade pip in our config.yaml: `pip install -U pip setuptools wheel twine` also upgrades `twine` using the **current/old** pip, hence it fails.

Fix: split the pip upgrade line into two.

.circleci/config.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           command: |
             python -m venv env
             . env/bin/activate
+            pip install -U pip
```
randomir, line 26: You can remove `pip` from the following line. I'd also move `setuptools` and `wheel` here, leaving `twine` only on the line below.

randomir (changes requested):

.circleci/config.yml
JoelPasvolsky: My thinking was that perhaps in the future one of the other dependencies might cause the same problem, so it would be better to first update pip then use the updated pip to do the rest.
Naturally I forgot to remove the pip from the split out line :-)

randomir: In the future we'll switch to `pyproject.toml`, so these will be updated automatically, but for now it's safe to do:
```
pip install -U pip setuptools wheel
pip install -U twine
```

randomir (approved):
LGTM!

PR #198, opened 2021-07-17
Update for dimod 0.10

arcondello:
Looks like the package is compatible, it was only the tests that broke. Still TODO for this PR:
- [ ] fix deprecation warnings
- [ ] add version tests to circleci


arcondello:
Duplicated by #199

PR #200, opened 2021-08-10
added new missing edge weight penalty for bidirectional graphs

zhemingfan:
Added a parameter for users to set for missing edge weight penalty in bidirectional TSP problems

### DIFF
--- a/dwave_networkx/algorithms/tsp.py
+++ b/dwave_networkx/algorithms/tsp.py
@@ -114,7 +114,7 @@ def traveling_salesperson(G, sampler=None, lagrange=None, weight='weight',
 traveling_salesman = traveling_salesperson
 
 
-def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
+def traveling_salesperson_qubo(G, lagrange=None, weight='weight', missing_edge_penalty='sum'):
     """Return the QUBO with ground states corresponding to a minimum TSP route.
 
     If :math:`|G|` is the number of nodes in the graph, the resulting qubo will have:
@@ -133,6 +133,9 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
 
     weight : optional (default 'weight')
         The name of the edge attribute containing the weight.
+    
+    missing_edge_penalty : number, optional (default 'sum')
+        For bi-directional graphs, the penalty associated with the back missing edges. Default is to use the sum all weights.
 
     Returns
     -------
@@ -154,6 +157,10 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
             lagrange = G.size(weight=weight)*G.number_of_nodes()/G.number_of_edges()
         else:
             lagrange = 2
+    
+    # default penalty format is sum
+    if missing_edge_penalty is "sum":
+        missing_edge_weight = sum(dict( (x[:-1], x[-1][weight]) for x in edges if weight in x[-1] ).values())
 
     # some input checking
     if N in (1, 2) or len(G.edges) != N*(N-1)//2:
--- a/dwave_networkx/algorithms/tsp.py
+++ b/dwave_networkx/algorithms/tsp.py
@@ -147,7 +147,7 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight', missing_edge_p
 
     """
     N = G.number_of_nodes()
-
+    print("hi")
     if lagrange is None:
         # If no lagrange parameter provided, set to 'average' tour length.
         # Usually a good estimate for a lagrange parameter is between 75-150%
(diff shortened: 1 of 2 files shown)


dwave_networkx/algorithms/tsp.py
```diff
@@ -154,7 +157,11 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
             lagrange = G.size(weight=weight)*G.number_of_nodes()/G.number_of_edges()
         else:
             lagrange = 2
-
+    
+    # default missing edge penalty format is sum
+    if missing_edge_penalty is "sum":
+        missing_edge_weight = sum(dict( (x[:-1], x[-1][weight]) for x in edges if weight in x[-1] ).values())
```
arcondello, line 163: I think `sum(weight for _, _, weight in G.edges.data('weight', default=0))` should also do it without needing to create an intermediate dict.

arcondello:
So steps remaining:

- [ ] Accept `missing_edge_penalty` as an input parameter. IMO it should default to `None` rather than `'sum'`. 
- [ ] Use this value in the construction in the case of a digraph. Probably as simple as changing `G[u][v][weight]` to `G[u][v].get('weight', missing_edge_weight)
- [ ] Add tests

dwave_networkx/algorithms/tsp.py
```diff
@@ -114,7 +114,7 @@ def traveling_salesperson(G, sampler=None, lagrange=None, weight='weight',
 traveling_salesman = traveling_salesperson
 
 
-def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
+def traveling_salesperson_qubo(G, lagrange=None, weight='weight', missing_edge_penalty='None'):
```
arcondello, line 117: I think this should be `missing_edge_weight=None`.

dwave_networkx/algorithms/tsp.py
```diff
@@ -154,10 +157,16 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
             lagrange = G.size(weight=weight)*G.number_of_nodes()/G.number_of_edges()
         else:
             lagrange = 2
+    
+    missing_edge_weight=""
+    
+    # default penalty format is sum
+    if missing_edge_penalty == "sum":
```
arcondello, line 164: And then this becomes 
```
if missing_edge_weight is None:
    missing_edge_weight = sum(weight for _, _, weight in G.edges.data('weight', default=0))
```

dwave_networkx/algorithms/tsp.py
```diff
@@ -133,6 +133,9 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
 
     weight : optional (default 'weight')
         The name of the edge attribute containing the weight.
+    
+    missing_edge_penalty : number, optional (default 'sum')
```
arcondello, line 137: This needs to be updated to match the function signature. Also should not be more than 80 characters wide.

dwave_networkx/algorithms/tsp.py
```diff
@@ -154,10 +157,16 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
...
+    
+    # default penalty format is sum
+    if missing_edge_penalty == "sum":
+        missing_edge_weight = sum(weight for _, _, weight in G.edges.data('weight', default=0))
 
     # some input checking
-    if N in (1, 2) or len(G.edges) != N*(N-1)//2:
-        msg = "graph must be a complete graph with at least 3 nodes or empty"
+    if N in (1, 2):
+        msg = "graph must be a complete graph"
```
arcondello, line 169: I believe this should be `msg = "graph must have at least 3 nodes or be empty"`

dwave_networkx/algorithms/tsp.py
JoelPasvolsky: @arcondello, didn't we recently increase to 100 chars?

tests/test_tsp.py
```diff
@@ -136,6 +136,41 @@ def test_k3(self):
 
         self.assertEqual(ground_count, len(min_routes))
 
+    def test_k3_bidirectional(self):
```
arcondello, line 139: So this tests the original interface, which is fine. But we should also add tests for `nx.Graph` with missing edges and `nx.MultiGraph` both with and without missing edges. Also test whether my user-specified `missing_edge_weight` gets used.

dwave_networkx/algorithms/tsp.py
arcondello: The strict limit is 100, though in doc strings IMO it's nice to keep it under 80. It's easier in docstrings since it's all text, so wrapping rarely makes it less readable.

PR #201, opened 2021-09-09
0.8.9 prerelease work

arcondello:


.circleci/config.yml
```diff
@@ -208,7 +208,7 @@ workflows:
           matrix:
             parameters:
               python-version: &python-versions [3.6.8, 3.7.9, 3.8.9, 3.9.4]
-              dimod-version: [==0.8.0, <0.10, '[preprocessing]<0.11']
+              dimod-version: [==0.8.0, <0.9, <0.10, <0.11]
```
randomir, line 211: Do we need 0.8.0?

randomir (approved):

.circleci/config.yml
arcondello: Eh, why not? It would be proactive work to remove it and that's currently the bottom of the range.

docs/bibliography.rst
```diff
@@ -9,4 +9,4 @@ Bibliography
 
 .. [FIA] Facchetti, G., Iacono G., and Altafini C. (2011). Computing global structural balance in large-scale signed social networks. PNAS, 108, no. 52, 20953-20958
 
-.. [DWMP] Dahl, E., "Programming the D-Wave: Map Coloring Problem:, https://www.dwavesys.com/sites/default/files/Map%20Coloring%20WP2.pdf
\ No newline at end of file
+.. [DWMP] Dahl, E., "Programming the D-Wave: Map Coloring Problem", https://www.dwavesys.com/media/htfgw5bk/map-coloring-wp2.pdf
```
randomir, line 12: Now, try guessing that link..

PR #202, opened 2021-09-09
Fix tag check

arcondello:


arcondello:
Went to check the blame for this one and it was me! :laughing:

randomir (approved):

Issue #203, opened 2021-10-01
DiGraph support


rjcortese:
Hello!

I want to use dwave-networkx with `networkx.DiGraph`s.
Specifically, I'm trying to pass a `networkx.DiGraph` to `dwave_networkx.traveling_salesperson` but it does not seem to work. My DiGraph is complete and I believe does meet other requirements.

Is there already support for DiGraph?
If not, I propose it be added.

Thanks!

arcondello:
We don't currently support it, though I don't think it would be especially difficult to make the change. https://github.com/dwavesystems/dwave-networkx/pull/200 actually does some of the work but I think that PR is abandoned and still needs a few features (see https://github.com/dwavesystems/dwave-networkx/pull/200#issuecomment-896192804). Would you be interested in making a PR?

rjcortese:
I can make a PR that addresses those issues, just need a bit of time to do it.

I was worried the QUBO formulation might be different from regular `Graph` QUBO formulation... but from looking at that PR, maybe that is not the case? :)

arcondello:
Yeah, there are two issues that need to be addressed:
1) Assign the appropriate edge weight based on the direction
2) Handle missing edges (if there are any)
but the formulation is pretty close.

Happy to answer questions if it would be helpful!

PR #205, opened 2021-10-05
Di graph support

rjcortese:
Hello,
This is for #203.
It picks up where #200 left off.
I think I've made necessary changes except for some finishing the tests... but I could use a review and feedback :)

### DIFF
--- a/dwave_networkx/algorithms/tsp.py
+++ b/dwave_networkx/algorithms/tsp.py
@@ -114,7 +114,7 @@ def traveling_salesperson(G, sampler=None, lagrange=None, weight='weight',
 traveling_salesman = traveling_salesperson
 
 
-def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
+def traveling_salesperson_qubo(G, lagrange=None, weight='weight', missing_edge_penalty='sum'):
     """Return the QUBO with ground states corresponding to a minimum TSP route.
 
     If :math:`|G|` is the number of nodes in the graph, the resulting qubo will have:
@@ -133,6 +133,9 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
 
     weight : optional (default 'weight')
         The name of the edge attribute containing the weight.
+    
+    missing_edge_penalty : number, optional (default 'sum')
+        For bi-directional graphs, the penalty associated with the back missing edges. Default is to use the sum all weights.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


dwave_networkx/algorithms/tsp.py
```diff
@@ -185,10 +195,10 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
             nextpos = (pos + 1) % N
 
             # going from u -> v
-            Q[((u, pos), (v, nextpos))] += G[u][v][weight]
+            Q[((u, pos), (v, nextpos))] += G[u][v].get(weight, missing_edge_weight)
 
             # going from v -> u
-            Q[((v, pos), (u, nextpos))] += G[u][v][weight]
+            Q[((v, pos), (u, nextpos))] += G[u][v].get(weight, missing_edge_weight)
```
arcondello, line 201: I think you want `G[v][u]` in this case, to allow different weights in different direction

tests/test_tsp.py
```diff
@@ -97,6 +97,29 @@ def test_start(self):
...
+            (1, 3, 2),
+            (3, 1, 2),
+            (2, 3, 2),
+            (3, 2, 1),
+        ])
+
+        route = dnx.traveling_salesperson(G, dimod.ExactSolver(), start=1)
+
+        self.assertEqual(len(route), len(G))
+        self.assertEqual(nx.path_weight(G, route, 'weight'), 4)
```
arcondello, line 120: Looks like this function requires a newer networkx than the minimum we support. I would be OK with bumping up the min networkx version in `setup.py`, but given that it's just for a test it might make sense to change the test a bit.

dwave_networkx/algorithms/tsp.py
```diff
@@ -185,10 +195,10 @@ def traveling_salesperson_qubo(G, lagrange=None, weight='weight'):
             nextpos = (pos + 1) % N
 
             # going from u -> v
-            Q[((u, pos), (v, nextpos))] += G[u][v][weight]
+            Q[((u, pos), (v, nextpos))] += G[u][v].get(weight, missing_edge_weight)
```
arcondello, line 198: I think you're going to need something like `G[u].get(v, {}).get(weight, missing_edge_weight)` since either the weight attribute or the edge itself may be missing.

arcondello:
Thanks for the PR! A few minor things but overall seems to be going in the right direction

rjcortese:
This is ready to merge now imo :)

arcondello (approved):

PR #209, opened 2021-11-02
Implement cqm partitioning algorithm and tests

necaisej:
Makes use of ExactCQMSolver (https://github.com/dwavesystems/dimod/pull/1037) in tests.


arcondello (commented):
Need to update the dimod version in setup.py https://github.com/dwavesystems/dwave-networkx/blob/7299ddf450c576a69357213760aceabf3d996fb7/setup.py#L33
and you need to change the tested versions in CI https://github.com/dwavesystems/dwave-networkx/blob/7299ddf450c576a69357213760aceabf3d996fb7/.circleci/config.yml#L211 to reflect the new minimum. So probably to `[==0.10.8, <0.11]`

arcondello:
And while you're at it, probably good to use the nicer `[==0.10.8, ~=0.10.0]` syntax.

arcondello:
Need to rebase against the main branch. I think the current `.circleci/config.yml`, `requirements.txt` and `setup.py` do not need to change.

dwave_networkx/algorithms/partition.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Reinterpret result as partition assignment over nodes
+    indicators = [key for key, value in possible_partitions.first.sample.items() if isclose(value, 1.)]
+    node_partition = {key[0]: key[1] for key in indicators}
+    
+    return node_partition
+    
+    
+def graph_partition_cqm(G, num_partitions):
+
```
arcondello, line 103: This needs a docstring.

dwave_networkx/algorithms/partition.py
```diff
@@ -0,0 +1,140 @@
...
+    response = sampler.sample_cqm(cqm, **sampler_args)
+
+    # Consider only results satisfying all constraints
+    possible_partitions = response.filter(lambda d: d.is_feasible)
+    
+    if not possible_partitions: 
+        raise Exception("No feasible solution could be found for this problem instance.")
+
+    # Reinterpret result as partition assignment over nodes
+    indicators = [key for key, value in possible_partitions.first.sample.items() if isclose(value, 1.)]
```
arcondello, line 96: this can be a generator rather than a list.

Issue #212, opened 2021-11-26
Does 'decorator==5.0.9' break 'dwave-networkx'?


ccalaza:
**Current Problem**
Currently at Forschungszentrum Juelich we are trying to integrate 'dwave-ocean-sdk==4.2.0' (which depends on 'dwave-networkx==0.8.10') into our HPC software stack. Since there is much more software installed in this environment this forces us to use the provided versions of software dependencies of 'dwave-networkx' rather than installing our own dependencies in order to avoid version collisions. 

Our HPC environment provides 'Python==3.9.6' and 'decorator==5.0.9', whereas the 'setup.py' of your package requires 'decorator>=4.1.0,<5.0.0'. As mentioned above, downgrading our version of 'decorator' would most likely affect many other Python packages downstream so we would like to avoid this. 

**Proposed Solution**
Would it be possible for you to check if using 'dwave-networkx' with 'decorator==5.0.9' instead of 'decorator<5' would cause any problems with your package? If you can confirm that this introduces no problems I can easily patch 'setup.py' and proceed with the installation without having to wait for a new version of your package to be released.

**Alternatives Considered**
Downgrading our 'decorator' version --> will potentially break other stuff on our side.
Waiting for a newer 'dwave-ocean-sdk' compatible with our 'decorator' version --> could delay our plans to launch JUNIQ and make our D-Wave system available to users.

Thanks in advance :)

arcondello:
I have tested it and there seem to be no issues with using a newer decorator. I will loosen or remove the requirement today, though it will take a little while for that change to propagate to the SDK.

ccalaza:
That's great news, thanks for the quick reply!

PR #213, opened 2021-11-26
Remove direct dependency on decorator package

arcondello:
Closes https://github.com/dwavesystems/dwave-networkx/issues/212

No longer depend on decorator directly. Newer NetworkX packages also remove their dependency so for those versions it won't be installed at all.

I did remove some of the (IMO) unneeded documentation, but I left the previous implementation more-or-less intact. Probably there is some refactoring that could be done there as well but if it isn't broken...


hhtong (approved):
LGTM

PR #214, opened 2021-12-07
Layout overhaul

boothby:
Normalizes node sizes across the board
Makes yield plots easier to read
Adds qubit-line plots

needs a little more documentation


boothby:
Caught a mistake; will patch later.  For now, have a preview.

yield2z100

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -153,10 +172,11 @@ def chimera_node_placer_2d(m, n, t, scale=1., center=None, dim=2):
     else:
         center = np.asarray(center)
 
-    paddims = dim - 2
-    if paddims < 0:
+    if dim < 2:
         raise ValueError("layout must have at least two dimensions")
 
+    paddims = np.zeros(dim - 2, dtype='float')
```
JoelPasvolsky, line 178: Perhaps add a comment above with a short explanation of what paddim is used for?

dwave_networkx/drawing/pegasus_layout.py
```diff
@@ -75,28 +80,27 @@ def pegasus_layout(G, scale=1., center=None, dim=2, crosses=False):
     if not isinstance(G, nx.Graph) or G.graph.get("family") != "pegasus":
         raise ValueError("G must be generated by dwave_networkx.pegasus_graph")
 
-    if G.graph.get('labels') == 'nice':
-        m = 3*(G.graph['rows']-1)
-        c_coords = chimera_node_placer_2d(m, m, 4, scale=scale, center=center, dim=dim)
-        def xy_coords(t, y, x, u, k):
-            return c_coords(3*y+2-t, 3*x+t, u, k)
+    labels = G.graph.get('labels')
+    xy_coords = pegasus_node_placer_2d(G, scale, center, dim, crosses=(crosses or labels == 'nice_coordinates'), normalize_kwargs = normalize_kwargs)
```
JoelPasvolsky, line 84: Maybe break into two code lines?

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -51,6 +51,11 @@ def chimera_layout(G, scale=1., center=None, dim=2):
         Number of dimensions. When dim > 2, all extra dimensions are
```
JoelPasvolsky, line 51: Now ``scale`` is divided by ``fabric_scale`` in the ``chimera_node_placer_2d`` function  - should the docstring for ``scale`` in line 44 be updated?

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -130,6 +135,11 @@ def chimera_node_placer_2d(m, n, t, scale=1., center=None, dim=2):
     dim : int (default 2)
```
JoelPasvolsky, line 135: Should the docstring for ``scale`` in line 129 be updated for line 163?

dwave_networkx/drawing/pegasus_layout.py
```diff
@@ -129,50 +133,71 @@ def pegasus_node_placer_2d(G, scale=1., center=None, dim=2, crosses=False):
     """
```
JoelPasvolsky, line 133: Should line 112 be updated for line 151?

dwave_networkx/drawing/pegasus_layout.py
```diff
@@ -129,50 +133,71 @@ def pegasus_node_placer_2d(G, scale=1., center=None, dim=2, crosses=False):
...
     tile_center = tile_width / 2 - .5
+    cross_shift = 2 if crosses else 0
 
     # want the enter plot to fill in [0, 1] when scale=1
-    scale /= m * tile_width
+    fabric_scale = m*tile_width - 2*odd_k_wobble - 1
+    scale /= fabric_scale
+
+    if normalize_kwargs is not None:
+        normalize_size_and_aspect(fabric_scale, 150, normalize_kwargs)
```
JoelPasvolsky, line 154: where does 150 come from?

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
         import matplotlib.pyplot as plt
         import matplotlib as mpl
     except ImportError:
-        raise ImportError("Matplotlib and numpy required for draw_chimera()")
+        raise ImportError("Matplotlib and numpy required for draw_yield()")
```
JoelPasvolsky, line 529: NumPy

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+            kwargs['width'] = 5*(fig_scale / scale)
+        else:
+            kwargs['width'] = 2*(fig_scale / scale)
+
+
+def draw_lineplot(G, layout, *, ax, node_size, width, nodelist=None,
+                  edgelist=None, node_color='blue', edge_color='black',
+                  cmap=None, vmin=None, vmax=None, edge_cmap=None,
+                  edge_vmin=None, edge_vmax=None, z_offset=0):
+    """Draws the graph G with line segments representing nodes
```
JoelPasvolsky, line 609: nodes.

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+    This function is meant to be a drop-in replacement for :func:`networkx.draw`
+    where nodes are associated with line segments (specified as 2x2 matrices
+    [[x0, y0], [x1, y1]]).  This function makes significant assumptions about
+    the edges of the graph G, that hold when G is a Chimera, Pegasus, or Zephyr
+    graph and the line segments are provided by :func:`chimera_layout`,
+    :func:`pegasus_layout` and :func:`zephyr_layout` respectively.  These graphs
+    have three classes of edges:
+
+        * internal edges between qubits whose line segments are perpendicular
+            and intersect at a point, which we draw with a circle located at the
```
JoelPasvolsky, line 620: For consistency: "which we draw with" --> "which are drawn with"

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+            point of intersection,
+        * external edges between qubits whose line segments are colinear, which
+            we draw as a line segment between the nearest endpoints, and
```
JoelPasvolsky, line 623: "which are drawn as"

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+        * odd edges between parallel qubits whose line segments are parallel and
+          overlap in a perpendicular projection, which we draw as a line segment
```
JoelPasvolsky, line 625: which are drawn as

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
-        if edgelist is None:
-            edgelist = G.edges() - faults_edgelist
+    Parameters
+    ----------
 
-        unused_node_color = [unused_color for v in nodelist]
-        unused_edge_color = [unused_color for v in edgelist]
+    G : networkx.Graph
+        A graph constructed by :func:`chimera_layout`, :func:`pegasus_layout, or
+        :func:`zephyr_layout`
```
JoelPasvolsky, line 633: :func:`zephyr_layout`.

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+    width : float
+        The width of line segments associated with edges, and half the width of
+        line segments associated with nodes.
+
+    nodelist : iterable or None (default=None)
+        The set of nodes to draw.  If None, all nodes from G are drawn.
+
+    edgelist : iterable or None (default=None)
+        The set of edges to draw.  If both nodelist and edgelist are None, all
+        edges of G are drawn.  If edgelist is None, all edges from the subgraph
```
JoelPasvolsky, line 654: ``edgelist ``
And elsewhere

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+    vmax : float or None (default=None)
+        Maximum value to use to use when normalizing node colors through cmap.
+
+    edge_cmap : string or matplotlib.ColorMap or None (default=None)
+        A colormap to color edges with.  Presumes that edge_color is a sequence
+        of floats.
+
+    edge_vmin : float or None (default=None)
+        Minimum value to use to use when normalizing edge colors through
+        edge_cmap
```
JoelPasvolsky, line 683: ``edge_cmap``.

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+
+    edge_vmax : float or None (default=None)
+        Maximum value to use to use when normalizing edge colors through
+        edge_cmap
```
JoelPasvolsky, line 687: As above

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+    
+    z_offset : int (default=0)
+        An offset to the zorder that various elements are drawn in.  Edge lines
```
JoelPasvolsky, line 690: Perhaps "zorder" for first usage before following explanation

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+        are drawn with zorder=z_offset; horizontal node lines are drawn with
+        zorder=zoffset+1; vertical node lines are drawn with zorder=zoffset+2,
+        and edge circles are drawn with zorder=zoffset+3.  This parameter can be
+        used to layer line plots over or under eachother.
```
JoelPasvolsky, line 694: eachother

dwave_networkx/drawing/zephyr_layout.py
```diff
@@ -52,6 +52,11 @@ def zephyr_layout(G, scale=1., center=None, dim=2):
         Number of dimensions. When dim > 2, all extra dimensions are
         set to 0.
 
+    normalize_kwargs : None or dict (default None)
+        A dict of keyword arguments to be used in a plotting function.  If not
+        None, we will populate the "ax" keyword with matplotlib axes, and the
+        "node_size" and "width" keywords with defaults if they are not set.
```
JoelPasvolsky, line 58: A link to ``draw_lineplot`` for the params would be helpful. Same for the others in Chimera & Pegasus.

JoelPasvolsky (commented):
Some comments from a first scan: if I have time, I'd like to try using the new functionality a bit and then returning to this.
Lovely addition!

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -141,10 +151,19 @@ def chimera_node_placer_2d(m, n, t, scale=1., center=None, dim=2):
     """
     import numpy as np
 
+    line_plot = False if normalize_kwargs is None else normalize_kwargs.get('line_plot')
```
JoelPasvolsky, line 154: In all the user functions (e.g. `draw_pegasus`) the `line_plot` is described as a Boolean but actually just its existence is checked so `dnx.draw_chimera(G, line_plot=3.14)` would work. Not sure if you want to enforce or not.

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,237 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+    Parameters
+    ----------
 
-        unused_node_color = [unused_color for v in nodelist]
-        unused_edge_color = [unused_color for v in edgelist]
+    G : networkx.Graph
+        A graph constructed by :func:`chimera_layout`, :func:`pegasus_layout, or
+        :func:`zephyr_layout`
+
+    pos : dict
```
JoelPasvolsky, line 635: I don't see a `pos` param in line 605 (`def draw_lineplot( ...`, I see a `layout` param

dwave_networkx/drawing/chimera_layout.py
boothby: There is no net change in behavior here; I just needed to capture the value of `fabric_scale` for use in `normalize_size_and_aspect`.

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -51,6 +51,11 @@ def chimera_layout(G, scale=1., center=None, dim=2):
         Number of dimensions. When dim > 2, all extra dimensions are
         set to 0.
 
+    normalize_kwargs : None or dict (default None)
+        A dict of keyword arguments to be used in a plotting function.  If not
+        None, we will populate the "ax" keyword with matplotlib axes, and the
+        "node_size" and "width" keywords with defaults if they are not set.
+
```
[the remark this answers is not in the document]
boothby:
```
    normalize_kwargs : None or dict (default None)
        A dict of keyword arguments to be used in a plotting function (see
        :func:`draw_chimera`, :func:`draw_chimera_embedding` or
        :func:`draw_yield`).  Populates the "ax" keyword with matplotlib axes,
        and the "node_size" and "width" keywords with defaults if they are not
        set.  If normalize_kwargs is None, do nothing.
```

Tense mismatch?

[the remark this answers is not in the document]
JoelPasvolsky: Shouldn't the link be to :func:`draw_lineplot`, that is where I found the params I could play with? 
Please use double back quotes (``) for params instead of double quotes (")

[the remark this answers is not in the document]
boothby: What happens is `normalize_kwargs["ax"]` gets set to an axis object -- it's an awkward case because it's both a keyword and a string.  Still think I should use double-backquotes?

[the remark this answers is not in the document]
boothby: I like this better, but I'll apply your advice to `normalize_size_and_aspect`
```
    normalize_kwargs : None or dict (default None)
        A dict of keyword arguments to be used in a plotting function (see
        :func:`networkx.draw` and :func:`draw_lineplot`).  Populates several
        keys in ``normalize_kwargs`` by calling :func:`normalize_size_and_aspect`
        with appropriate parameters for G. If normalize_kwargs is None, do 
        nothing.
```

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -209,6 +240,14 @@ def draw_chimera(G, **kwargs):
         form {edge: bias, ...}. Each bias should be numeric. Self-loop
         edges (i.e., :math:`i=j`) are treated as linear biases.
 
+    line_plot : boolean (optional, default False)
+        If line_plot is True, then qubits are drawn as line segments, and edges
+        are drawn either as line segments between qubits, or as circles where
+        two qubits overlap.  In this drawing style, the interpretation the width
```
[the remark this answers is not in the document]
boothby:
```
    line_plot : boolean (optional, default False)
        If line_plot is True, then qubits are drawn as line segments, and edges
        are drawn either as line segments between qubits, or as circles where
        two qubits overlap.  In this drawing style, the interpretation of the
        ``width`` and ``node_size`` parameters (provided in ``kwargs``)
        determine the area of the circles and line widths respectively.  See
        :func:`dwave_networkx.qubit_layout.draw_lineplot` for more information.
```

dwave_networkx/drawing/chimera_layout.py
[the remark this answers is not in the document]
JoelPasvolsky: "Populates several keys in ``normalize_kwargs``" is a bit off: the dict isn't populated here, it can be set here by the user, and if used, the called function sets default values for others. So probably be clearer if the 2nd sentence said something like "Optional keywords that are not provided are set to default values by the :func:`normalize_size_and_aspect` function" 

BTW, is "normalize" clearer in this context than something like "matplot_kwargs" or "graph_kwargs" or "plot_kwargs"? If I just saw the keyword I would not connect it to making the graphing eye friendly

dwave_networkx/drawing/pegasus_layout.py
boothby: It's a tight upper bound on the entries of lines 181 and 183; can you recommend clearer language for line 149?

dwave_networkx/drawing/chimera_layout.py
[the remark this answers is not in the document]
JoelPasvolsky: Yes. Just I think "provided in ``kwargs``" might be ``normalize_kwargs`, right?

dwave_networkx/drawing/pegasus_layout.py
boothby: It's a magic value, determined experimentally.

JoelPasvolsky: Is it worth defining it as a constant above line 153 with a comment and then using that here?

dwave_networkx/drawing/chimera_layout.py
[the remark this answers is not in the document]
boothby: Eh, "plot_kwargs" is fewer characters.  Yeah, I hated that name.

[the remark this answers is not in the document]
boothby:
```
    plot_kwargs : None or dict (default None)
        A dict of keyword arguments to be used in a plotting function (see
        :func:`networkx.draw` and :func:`draw_lineplot`), to scale node and edge
        sizes appropriately to the graph `G`.  Optional keys in ``plot_kwargs``
        are set to default values by calling :func:`normalize_size_and_aspect`
        with appropriate parameters for G. Do nothing if ``plot_kwargs`` is None
```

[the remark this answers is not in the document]
JoelPasvolsky: You could also mention that you added coloring capability. Maybe something like

plot_kwargs : None or dict (default None)
        Parameters to shape and color plotting, as a dict of keyword arguments (see
        :func:`networkx.draw` and :func:`draw_lineplot`).  Unspecified parameters of
        ``plot_kwargs`` are set to default values appropriate to the graph by the underlying
        :func:`normalize_size_and_aspect` call. Does nothing if ``plot_kwargs`` is None.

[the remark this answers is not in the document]
boothby: I see now that I have not documented `normalize_size_and_aspect`.  I don't do anything with coloring there -- what's getting normalized is
 1. if an axis (`"ax"`) is not set, then I create one
 2. the aspect ratio of the axis is set to 1
 3. axes are turned off (this is confusing because of the overloaded term "axis" -- here I mean that we will not draw lines and tick marks for the x- and y-axes on the plot)
 4. if `"node_size"` and/or `"width"` are not set, they are calculated from the fabric scale, figure scale, and magic numbers

So, in steps (1-3), the matplotlib `"axis"` object is being normalized.  In step 4, the feature sizes in the drawings are being normalized.

dwave_networkx/drawing/pegasus_layout.py
boothby: I don't think so

dwave_networkx/drawing/chimera_layout.py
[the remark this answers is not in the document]
boothby: No, `normalize_kwargs` does not appear in that argument list.

dwave_networkx/drawing/qubit_layout.py
boothby: so aggressive

dwave_networkx/drawing/qubit_layout.py
boothby: Really?  You've been telling me to put periods on the ends of these things.

dwave_networkx/drawing/qubit_layout.py
boothby: The offset is absolute, and set per lineplot.  This functionality is used by `draw_qubit_embedding`, where we first draw the fabric of nodes and edges which are not used by the embedding and then draw nodes and edges used by the embedding -- the unused pieves have a lower `z_offset` so that they are drawn beneath the used portions.

dwave_networkx/drawing/chimera_layout.py
boothby: I do not want to enforce that.  If a user puts a number into a boolean field, they should expect that 0 will be treated as False and anything else should be treated as True.  In the case they have omitted `line_plot`, then the default value of `None` will be returned from `get('line_plot')`, which is in turn interpreted as `False`.

dwave_networkx/drawing/qubit_layout.py
boothby: yes, thank you

dwave_networkx/drawing/qubit_layout.py
JoelPasvolsky: :-) "each other."

boothby: ahhhh that makes sense :D

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -51,6 +51,13 @@ def chimera_layout(G, scale=1., center=None, dim=2):
         Number of dimensions. When dim > 2, all extra dimensions are
         set to 0.
 
+    plot_kwargs : None or dict (default None)
+        A dict of keyword arguments to be used in a plotting function (see
+        :func:`networkx.draw` and :func:`draw_lineplot`), to scale node and edge
+        sizes appropriately to the graph `G`.  Optional keys in ``plot_kwargs``
+        are set to default values by the :func:`normalize_size_and_aspect`
```
JoelPasvolsky, line 58: ":func:`normalize_size_and_aspect`" --> two issues:
1. Needs to be something like :func:`~dwave_networkx.drawing.qubit_layout.normalize_size_and_aspect`
2. You can only link to this function once you add it to an RST file so it's rendered to the docs.

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -97,15 +104,15 @@ def chimera_layout(G, scale=1., center=None, dim=2):
...
+        xy_coords = chimera_node_placer_2d(m, n, t, scale, center, dim, plot_kwargs=plot_kwargs)
 
         # compute our coordinates
         pos = {v: xy_coords(i, j, u, k) for v, (i, j, u, k) in chimera_indices.items()}
 
     return pos
 
 
-def chimera_node_placer_2d(m, n, t, scale=1., center=None, dim=2):
+def chimera_node_placer_2d(m, n, t, scale=1., center=None, dim=2, plot_kwargs=None):
```
JoelPasvolsky, line 115: This functionality needs to be added to https://docs.ocean.dwavesys.com/en/stable/docs_dnx/reference/drawing.html#module-dwave_networkx.drawing.chimera_layout (doc/reference/drawing.rst): 

```
.. autosummary::
   :toctree: generated/

   chimera_layout
   chimera_node_placer_2d
   draw_chimera
```

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -131,6 +138,13 @@ def chimera_node_placer_2d(m, n, t, scale=1., center=None, dim=2):
         Number of dimensions. When dim > 2, all extra dimensions are
         set to 0.
 
+    plot_kwargs : None or dict (default None)
```
JoelPasvolsky, line 141: Same comments as in def chimera_layout() above

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -209,6 +246,14 @@ def draw_chimera(G, **kwargs):
         form {edge: bias, ...}. Each bias should be numeric. Self-loop
         edges (i.e., :math:`i=j`) are treated as linear biases.
 
+    line_plot : boolean (optional, default False)
+        If ``line_plot`` is True, then qubits are drawn as line segments, and
+        edges are drawn either as line segments between qubits, or as circles
+        where two qubits overlap.  In this drawing style, the interpretation of
+        the ``width`` and ``node_size`` parameters (provided in ``kwargs``)
+        determine the area of the circles and line widths respectively.  See
+        :func:`dwave_networkx.qubit_layout.draw_lineplot` for more information.
```
JoelPasvolsky, line 255: Prefixing with a tilde (~) will be shorter, it might need a ``.drawing.`` before the func name, and it needs to be rendered in the docs first. 

Also appears in funcs below.

dwave_networkx/drawing/pegasus_layout.py
```diff
@@ -58,6 +58,13 @@ def pegasus_layout(G, scale=1., center=None, dim=2, crosses=False):
         rather than L configuration. Ignored if G is defined with
         ``nice_coordinates=True``.
 
+    plot_kwargs : None or dict (default None)
```
JoelPasvolsky, line 61: Same comments as in def chimera_layout() above

dwave_networkx/drawing/pegasus_layout.py
```diff
@@ -120,6 +129,13 @@ def pegasus_node_placer_2d(G, scale=1., center=None, dim=2, crosses=False):
         If True, :math:`K_{4,4}` subgraphs are shown in a cross
         rather than L configuration.
 
+    plot_kwargs : None or dict (default None)
```
JoelPasvolsky, line 132: Same comments as in def chimera_layout() above

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -23,13 +23,16 @@
 from networkx import draw
 
 from dwave_networkx.drawing.distinguishable_colors import distinguishable_color_map
+from itertools import repeat, chain
+
+from numbers import Number
 
 __all__ = ['draw_qubit_graph']
```
JoelPasvolsky, line 30: What about funcs ``draw_embedding``, ``draw_yield``, ``normalize_size_and_aspect``, ``draw_lineplot``? Shouldn't these be added? And added to the appropriate RST file?

dwave_networkx/drawing/zephyr_layout.py
```diff
@@ -52,6 +52,13 @@ def zephyr_layout(G, scale=1., center=None, dim=2):
         Number of dimensions. When dim > 2, all extra dimensions are
         set to 0.
 
+    plot_kwargs : None or dict (default None)
```
JoelPasvolsky, line 55: Same as above comments on func pathnames

JoelPasvolsky (commented):
Will you be adding the https://github.com/dwavesystems/dwave-networkx/blob/main/dwave_networkx/drawing/qubit_layout.py functions to RST files (see comment below on ``__all__ = ['draw_qubit_graph']``)? If so, I'll render those docstrings and see how they look too.

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -468,34 +526,270 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+        try:
+            import matplotlib.pyplot as plt
+        except ImportError:
+            raise ImportError("Matplotlib required for graph drawing")
+        cf = plt.gcf()
+    else:
+        cf = ax.get_figure()
+    cf.set_facecolor("w")
+    if ax is None:
+        if cf._axstack() is None:
```
JoelPasvolsky, line 612: @boothby, I started this morning but am running into broken functionality:
 ```
>>> import dwave_networkx as dnx
>>> G = dnx.chimera_graph(1)
>>> dnx.draw_chimera(G)
```
fails on ``TypeError: '_AxesStack' object is not callable``. 
From what I see, this line is not trigger by ``dnx.chimera_layout(G)`` but will be when that method has `kwargs`, which it will have when called through ``chimera_graph()``.

Also, I see that, 
```
>>> import matplotlib.pyplot as plt
>>> f = plt.figure()   # equivalent to cf = plt.gcf() in line 607
>>> f.__axstack()
```
fails on ``AttributeError: 'Figure' object has no attribute '__axstack'``

My venv uses ``matplotlib`` version 3.6.3. Is ``__axstack()`` supported in some previous version? This repo has no version requirement on matplotlib (or requirement that it be installed at all)

File: dwave_networkx/drawing/qubit_layout.py [hunk 32]
```diff
@@ -23,13 +23,16 @@
 from networkx import draw
```
JoelPasvolsky, line 24: [line 16] "Tools to visualize Chimera lattices ..." --> change to "Tools to visualize QPU lattices ..."

JoelPasvolsky: Should we not add a ``from dwave_networkx.drawing.qubit_layout import *`` line to the [init file](https://github.com/dwavesystems/dwave-networkx/blob/main/dwave_networkx/drawing/__init__.py)?

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -60,6 +63,16 @@ def draw_qubit_graph(G, layout, linear_biases={}, quadratic_biases={},
         be. If not provided, the colormap will default to the middle of
         min/max values provided.
 
+    line_plot : boolean (optional, default False)
```
JoelPasvolsky, line 66: I'm encountering some problems with this parameter: 

1. "In this drawing style, the interpretation the width and node_size parameters (provided in kwargs) determines..." 

I think anything in ``kwargs`` should be optional, but not providing ``width`` and ``node_size`` parameters will cause an error: ``TypeError: draw_lineplot() missing 2 required keyword-only arguments: 'node_size' and 'width'``

2. The function fails for what seems to me simple input:
```
>>> import dwave_networkx as dnx
>>> G = dnx.chimera_graph(1)
>>> layout = dnx.chimera_layout(G)
>>> draw_qubit_graph(G, layout, line_plot=True, node_size=2, width=2)
```
Fails with ``TypeError: cannot unpack non-iterable numpy.float64 object``

boothby: This function isn't really meant to be called by users directly (note that it is not in the root `dwave_networkx` namespace).  The following should work:

```python
import dwave_networkx as dnx
G = dnx.chimera_graph(1)
layout = dnx.chimera_layout(G, plot_kwargs={'line_plot': True})
dnx.drawing.qubit_layout.draw_qubit_graph(G, layout, line_plot=True, node_size=2, width=2)
```

JoelPasvolsky: This still fails but with the problem [noted here](https://github.com/dwavesystems/dwave-networkx/pull/214#discussion_r1089351957):

```
>>> dnx.chimera_layout(G, plot_kwargs={'line_plot': True})
...
TypeError: '_AxesStack' object is not callable
```

dwave_networkx/drawing/qubit_layout.py
boothby: No, I don't think we should.  These functions are for internal use.

dwave_networkx/drawing/qubit_layout.py
boothby: yeah, I'm working on that

boothby: If you'd like; try it with matplotlib 3.4.1

dwave_networkx/drawing/qubit_layout.py
boothby: Identical to https://github.com/networkx/networkx/pull/5937; this has an easy fix (thanks, git-blame!)

dwave_networkx/drawing/qubit_layout.py
JoelPasvolsky: yes, that works, thanks

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -442,22 +487,35 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+        If line_plot is True, then qubits are drawn as line segments, and edges
+        are drawn either as line segments between qubits, or as circles where
+        two qubits overlap.  In this drawing style, the interpretation the width
+        and node_size parameters (provided in kwargs) determines the area of the
+        circles, and line widths, respectively.  Qubit line segments are given
+        twice the width of edges.  Layout should be a dict of the form
+        {node: ((x0, y0), (y0, x0)), ...} -- instead of coordinates, the nodes
+        are associated with endpoints of n-dimensional line segments.
```
JoelPasvolsky, line 517:
```suggestion
        If ``line_plot`` is ``True``, qubits are drawn as line segments and edges
        are drawn as either line segments between qubits or circles where
        two qubits overlap.  In this drawing style, the interpretation the ``width``
        and ``node_size`` parameters (provided in ``kwargs``) determines the area of the
        circles, and line widths, respectively.  Qubit line segments are given
        twice the width of edges.  Layout should be a dict of the form
        ``{node: ((x0, y0), (y0, x0)), ...}``; instead of coordinates, the nodes
        are associated with endpoints of n-dimensional line segments.
```

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -442,22 +487,35 @@ def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
...
+
     kwargs : optional keywords
        See networkx.draw_networkx() for a description of optional keywords,
        with the exception of the `pos` parameter which is not used by this
```
JoelPasvolsky, line 521:
```suggestion
       with the exception of the ``pos`` parameter which is not used by this
       function. If ``linear_biases`` or ``quadratic_biases`` are provided,
       any provided ``node_color`` or ``edge_color`` arguments are ignored.
       
       Examples
       --------
    
       This example illustrates how functions such as :func:`.draw_pegasus` can make use of this internal function.
    
       .. code-block:: python
    
           import dwave_networkx as dnx
           import matplotlib.pyplot as plt
           from dwave_networkx.drawing.qubit_layout import draw_yield
           G = dnx.chimera_graph(1)
           layout = dnx.chimera_layout(G)
           draw_yield(G.subgraph([0, 4, 5, 6, 7]), layout, G, unused_color="green", fault_color="red")
   """
```

This one will need a minor edit for indent spaces and cleaning up the three lines below after being committed.

JoelPasvolsky (commented):
I still have two functions remaining that I'll do right after this but I'm submitting these proposed updates now because (1) I want to submit a separate comment on what looks like a bug and (2) I have a lot of work that would be lost if this review hits a snag

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -424,9 +469,9 @@ def unoverlapped_embedding(G, emb, interaction_edges):
 
 
 def draw_yield(G, layout, perfect_graph, unused_color=(0.9,0.9,0.9,1.0),
```
JoelPasvolsky, line 471: @boothby, I'm seeing something weird with this function though it might be because I'm using it directly rather than through the public functions as intended, where it might not affect anything. The following code snippet, 

```
import dwave_networkx as dnx
import matplotlib.pyplot as plt
from dwave_networkx.drawing.qubit_layout import draw_yield
G = dnx.chimera_graph(1)
layout = dnx.chimera_layout(G)
draw_yield(G.subgraph([0, 4, 5, 6, 7]), layout, G, unused_color="green", fault_color="red")
```

generates the nodes with/without faults correctly but the edges are offset from the nodes:

image

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -240,6 +259,16 @@ def draw_embedding(G, layout, emb, embedded_graph=None, interaction_edges=None,
...
+    line_plot : boolean (optional, default False)
+        If line_plot is True, then qubits are drawn as line segments, and edges
+        are drawn either as line segments between qubits, or as circles where
+        two qubits overlap.  In this drawing style, the interpretation the width
+        and node_size parameters (provided in kwargs) determines the area of the
+        circles, and line widths, respectively.  Qubit line segments are given
+        twice the width of edges.  Layout should be a dict of the form
+        {node: ((x0, y0), (y0, x0)), ...} -- instead of coordinates, the nodes
+        are associated with endpoints of n-dimensional line segments.
+
```
JoelPasvolsky, line 271:
```suggestion

    kwargs : optional keywords
       See :func:`networkx.draw_networkx` for a description of optional keywords,
       with the exception of the ``pos`` parameter which is not used by this
       function. If ``linear_biases`` or ``quadratic_biases`` are provided,
       any provided ``node_color`` or ``edge_color`` arguments are ignored.

    Examples
    --------
    
    This example illustrates how functions such as :func:`.draw_pegasus` can make use of this internal function.
    
    .. code-block:: python
    
        import dwave_networkx as dnx
        import matplotlib.pyplot as plt
        from dwave_networkx.drawing.qubit_layout import draw_embedding
        G = dnx.chimera_graph(1)
        layout = dnx.chimera_layout(G)
        draw_embedding(G, layout, {0:[0, 4, 5], 1: [1, 6], 2: [2, 6], 3: [3, 7]})

    """
```

This one will need the deletion of lines 272-277 after committing.

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -240,6 +259,16 @@ def draw_embedding(G, layout, emb, embedded_graph=None, interaction_edges=None,
...
 
+    line_plot : boolean (optional, default False)
+        If line_plot is True, then qubits are drawn as line segments, and edges
+        are drawn either as line segments between qubits, or as circles where
+        two qubits overlap.  In this drawing style, the interpretation the width
+        and node_size parameters (provided in kwargs) determines the area of the
+        circles, and line widths, respectively.  Qubit line segments are given
+        twice the width of edges.  Layout should be a dict of the form
+        {node: ((x0, y0), (y0, x0)), ...} -- instead of coordinates, the nodes
+        are associated with endpoints of n-dimensional line segments.
```
JoelPasvolsky, line 270:
```suggestion
    line_plot : boolean (optional, default False)
        If ``line_plot`` is ``True``, qubits are drawn as line segments and edges
        are drawn as either line segments between qubits or circles where
        two qubits overlap.  In this drawing style, the interpretation of the ``width``
        and ``node_size`` parameters (provided in ``kwargs``) determines the area of the
        circles, and line widths, respectively.  Qubit line segments are given
        twice the width of edges.  Layout should be a dict of the form
        ``{node: ((x0, y0), (y0, x0)), ...}``; instead of coordinates, the nodes
        are associated with endpoints of n-dimensional line segments.
```

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -60,6 +63,16 @@ def draw_qubit_graph(G, layout, linear_biases={}, quadratic_biases={},
...
```
JoelPasvolsky, line 74:
```suggestion
        If ``line_plot`` is ``True``, qubits are drawn as line segments and edges
        are drawn as either line segments between qubits or circles where
        two qubits overlap. In this drawing style, the interpretation of the ``width``
        and ``node_size`` parameters (provided in ``kwargs``) determines the area of the
        circles, and line widths, respectively.  Qubit line segments are given
        twice the width of edges.  Layout should be a dict of the form
        ``{node: ((x0, y0), (y0, x0)), ...}``; instead of coordinates, the nodes
        are associated with endpoints of n-dimensional line segments.
```

JoelPasvolsky (commented):
Okay, these are great functionality additions. 
There are still a few small changes (mostly formatting and an example) I'd like but because they're not on code lines directly updated in this PR it's messy to propose them here---I can submit a followup PR to you once this is merged.

dwave_networkx/drawing/qubit_layout.py
boothby: I'm not reproducing that on my end (matplotlib 3.4.1, networkx 2.4).  Can you try the public interface?

JoelPasvolsky: Okay, I finally managed to reproduce it through the public interface:

1. It does not happen for matplotlib 3.4.1, networkx 2.4; it happens if you upgrade to matplotlib 3.6.3 (dwave-netwokx doesn't set a version requirement for Matplotlib so users might have more recent versions)
2. We need to add ``draw_chimera_yield`` and ``draw_pegasus_yield`` to [the docs](https://docs.ocean.dwavesys.com/en/stable/docs_dnx/reference/drawing.html), currently there's only ``draw_zephyr_yield``

```
import dwave_networkx as dnx
import matplotlib.pyplot as plt
from dwave_networkx.drawing.qubit_layout import draw_embedding
G = dnx.chimera_graph(1)
layout = dnx.chimera_layout(G)
fig, ax = plt.subplots(figsize=(10, 5))
dnx.draw_chimera_yield(G.subgraph([0, 4, 5, 6, 7]), unused_color="green", fault_color="red", ax=ax)
plt.show()
```

Shows up once you zoom in as: 

image

PR #227, opened 2023-04-21
Update drawing examples for fixed overlay

JoelPasvolsky:
Something changed (networkx or matplotlib?) since I created these examples and the 2nd no longer overlays the first (scale mismatch). I'll check if the problem exists in other places too. The updates below produce the expected output:

image
image

CC @davage


arcondello (approved):
LGTM

PR #230, opened 2023-05-09
Refresh CI etc

arcondello:
I just test the min and max of the dependency range rather than all the intermediate version. Given the number of dependencies this seems simpler.


randomir (approved):
LGTM!

arcondello:
Turns out that it's silently failing the osx tests
image
will fix

PR #231, opened 2023-05-09
Explicitly specify osx resource class in CI

arcondello:
It's supposed to route [automatically](https://discuss.circleci.com/t/macos-resource-deprecation-update/46891), but currently we get an error
image
so specifying it explicitly.

See https://github.com/dwavesystems/dwave-networkx/pull/230#issuecomment-1540586504


randomir (approved):
I guess they're not automatically upgrading to `gen2` because we're not on the Free plan.

Issue #232, opened 2023-11-16
draw_topology functions error when linear xor quadratic biases is provided


kevinchern:
**Description**
draw_topology functions, e.g., `draw_pegasus`, errors when you supply linear biases and not quadratic biases (vice versa too)

**To Reproduce**
```
import dwave_networkx as dnx
G = dnx.pegasus_graph(3)
# This case works as expected
dnx.draw_pegasus(G,  linear_biases={117:-1}, quadratic_biases={(128, 130): -1})
# This case errors out
try:
    dnx.draw_pegasus(G, quadratic_biases={(128, 130): -1})
except Exception as e:
    print(e)
# This case errors out
try:
    dnx.draw_pegasus(G, linear_biases={117:-1})
except Exception as e:
    print(e)
```

**Additional context**
Culprit: https://github.com/dwavesystems/dwave-networkx/blob/0.8.14/dwave_networkx/drawing/qubit_layout.py#L117C1-L120

PR #233, opened 2023-11-16
Fix undefined variables for color maps

kevinchern:
For this issue https://github.com/dwavesystems/dwave-networkx/issues/232

### DIFF
--- a/dwave_networkx/drawing/qubit_layout.py
+++ b/dwave_networkx/drawing/qubit_layout.py
@@ -131,61 +131,44 @@ def draw_qubit_graph(G, layout, linear_biases=None, quadratic_biases=None,
 
         # since we're applying the colormap here, matplotlib throws warnings if
         # we provide these arguments and it doesn't use them.
-        cmap = kwargs.pop('cmap', plt.get_cmap('coolwarm'))
-        vmin = kwargs.pop('vmin', -1 * vmag)
-        vmax = kwargs.pop('vmax', vmag)
+        cmap = kwargs.pop('cmap', None) or plt.get_cmap('coolwarm')
+        vmin = kwargs.pop('vmin', None) or -1 * vmag
+        vmax = kwargs.pop('vmax', None) or vmag
 
-        edge_cmap = kwargs.pop('edge_cmap', plt.get_cmap('coolwarm'))
-        edge_vmin = kwargs.pop('edge_vmin', -1 * vmag)
-        edge_vmax = kwargs.pop('edge_vmax', vmag)
+        edge_cmap = kwargs.pop('edge_cmap', None) or plt.get_cmap('coolwarm')
+        edge_vmin = kwargs.pop('edge_vmin', None) or -1 * vmag
+        edge_vmax = kwargs.pop('edge_vmax', None) or vmag
 
         if linear_biases and quadratic_biases:
-            global_vmin = min(edge_vmin, vmin)
-            global_vmax = max(edge_vmax, vmax)
+            final_vmin = min(edge_vmin, vmin)
+            final_vmax = max(edge_vmax, vmax)
+            mapper = cmap
 
-            if midpoint is None:
-                midpoint = (global_vmax + global_vmin) / 2.0
-            norm_map = mpl.colors.TwoSlopeNorm(midpoint, vmin=global_vmin, vmax=global_vmax)
-
-            node_color = [cmap(norm_map(node)) for node in node_color]
-            edge_color = [cmap(norm_map(edge)) for edge in edge_color]
-            mpl.colorbar.ColorbarBase(cax, cmap=cmap, norm=norm_map, orientation='vertical')
-
-        # if the biases are provided, then add a legend explaining the color map
         elif linear_biases:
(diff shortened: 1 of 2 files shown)


boothby (approved):

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -143,17 +125,16 @@ def node_color(v):
...
-            vmin = -1 * vmag
-
-        if vmax is None:
-            vmax = vmag
 
-        if edge_vmin is None:
-            edge_vmin = -1 * vmag
+        # since we're applying the colormap here, matplotlib throws warnings if
+        # we provide these arguments and it doesn't use them.
+        cmap = kwargs.pop('cmap', plt.get_cmap('coolwarm'))
```
randomir, line 131: Is a change in semantics intended here? Consider the case when user supplies `cmap=None`.

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -143,17 +125,16 @@ def node_color(v):
...
+        vmin = kwargs.pop('vmin', -1 * vmag)
+        vmax = kwargs.pop('vmax', vmag)
 
-        if edge_vmax is None:
-            edge_vmax = vmag
+        edge_cmap = kwargs.pop('edge_cmap', plt.get_cmap('coolwarm'))
+        edge_vmin = kwargs.pop('edge_vmin', -1 * vmag)
+        edge_vmax = kwargs.pop('edge_vmax', vmag)
```
randomir, line 137: Actually, the same question applies to all of these.

randomir (commented):
Since you're already touching `linear_biases`/`quadratic_biases` here, would you mind fixing their [mutable default argument](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)?

Also, adding (at least some smoke) tests for `draw_qubit_graph` that verify the fix would be nice.

dwave_networkx/drawing/qubit_layout.py
kevinchern: The semantics should be the same. In the previous commit, we have
```
if linear_biases:
    cmap = kwargs.pop('cmap', None)
...
if cmap is None:
    cmap = plt.get_cmap('coolwarm')
```
which would have the same behaviour. The issue is if `linear_biases` was never supplied, then the second conditional throws an error as `cmap` is undefined.

kevinchern:
I'll put in some tests

dwave_networkx/drawing/qubit_layout.py
randomir: No, previously if the function was called with `cmap=None` (explicitly set), you would end up with `cmap = plt.get_cmap('coolwarm')` (assuming it doesn't fail).

Now, if user explicitly sets `cmap=None`, the final value will be `cmap = None`.

kevinchern: AH, OK I misunderstood.

dwave_networkx/drawing/qubit_layout.py
kevinchern: Fixed with https://github.com/dwavesystems/dwave-networkx/pull/233/files#diff-344ce7a800a7c461ff087cf0103c3ba81b6af1935497c0b179431a811faee391R134-R140

tests/test_qubit_layout.py
```diff
@@ -0,0 +1,59 @@
...
+    import numpy as np
+except ImportError:
+    np = False
+
+_display = os.environ.get('DISPLAY', '') != ''
+
+
+@unittest.skipUnless(np and plt, "No numpy or matplotlib")
+class TestDrawing(unittest.TestCase):
+    @unittest.skipUnless(_display, " No display found")
```
randomir, line 35: Not sure you need this if using non-interactive matplotlib backend. Btw, which backend is used here? Hopefully not interactive..

dwave_networkx/drawing/qubit_layout.py
```diff
@@ -143,47 +128,37 @@ def node_color(v):
...
-
-        if vmax is None:
-            vmax = vmag
 
-        if edge_vmin is None:
-            edge_vmin = -1 * vmag
+        # since we're applying the colormap here, matplotlib throws warnings if
+        # we provide these arguments and it doesn't use them.
+        cmap = kwargs.pop('cmap', None) or plt.get_cmap('coolwarm')
+        vmin = kwargs.pop('vmin', None) or -1 * vmag
```
randomir, line 135: `or` might not be what you want here. E.g. if user explicitly sets `vmin=0`, you override it to `-vmag`.

randomir: When I say "here", I mean in this line, and a few other lines below.

tests/test_qubit_layout.py
kevinchern: I added this for consistency with existing tests, e.g., https://github.com/dwavesystems/dwave-networkx/blob/main/tests/test_pegasus_layout.py#L68-L104

PR #236, opened 2024-05-22
Zephyr layout coords fix

mahdiehmalekian:
A bug in `zephyr_layout` in the file `dwave-networkx/drawing/zephyr_layout.py` is fixed so that if ``scale`` = scale, all positions fit within [0, scale] on the x-axis and [-scale, 0] on the y-axis.

A couple of tests are added to `tests/test_zephyr_layout.py` that would fail before fixing the bug. These test the range of x value and y-value produced by `zephyr_layout` for the default scale=1 and scale=10.

Similarly, for `pegasus_layout` and `test_pegasus_layout.py`


dwave_networkx/drawing/zephyr_layout.py
```diff
@@ -139,9 +157,8 @@ def _xy_coords(u, w, k, j, z):
         else:
             xy = np.array([W, -Z])
 
-
         return np.hstack((xy * scale, np.zeros(paddims))) + center
-
+
```
jackraymond, line 161:
```suggestion

```

jackraymond:
@randomir - can you assign myself and @AndyZzzZzzZzz as reviewers?

tests/test_zephyr_layout.py
```diff
@@ -73,6 +73,39 @@ def test_zephyr_layout_coords(self):
...
+        self.assertAlmostEqual(max_x, 10, delta=1e-5, msg="max_x should be approximately 10")
+        
+    def test_zephyr_layout_yrange(self):
+        G = dnx.zephyr_graph(2, 4)
+        pos = dnx.zephyr_layout(G, scale=10)
+        y_coords = [val[1] for val in pos.values()]
+        min_y, max_y = min(y_coords), max(y_coords)
+        self.assertAlmostEqual(min_y, -10, delta=1e-5, msg="min_y should be approximately -10")
+        self.assertAlmostEqual(max_y, 0, delta=1e-5, msg="max_y should be approximately 0")
+
```
AndyZzzZzzZzz, line 108:
```suggestion
```

mahdiehmalekian:
@randomir Is this ready to be merged?

jackraymond (changes requested):
I understand it would be standard if numpy was imported at the top rather than within the layout function. Double check with @randomir  

range != [0,1] in layout as described. The update corrects this problem, unit tests are good. 

It is not clear to me why scale=1 must imply [0,1], when I use pos in combination with draw_networkx (for example) plots look equally good before and after this pull request (to my eye, superficially). Is there a specific example of when this causes unexpected behaviour? Happy to assume there is one and approve the request though.

mahdiehmalekian:
You're correct that the plots don't look different before and after this PR. However the document says scale=1 gives values in [0,1], and this is a feature that was relied on in the layout algorithm (in `dnx_layout` in `layout.py`) and the bug in this code caused the layout algorithm to give poor results.

mahdiehmalekian:
@randomir
Could you advise on numpy import?

dwave_networkx/drawing/zephyr_layout.py
```diff
@@ -78,7 +79,24 @@ def zephyr_layout(G, scale=1., center=None, dim=2):
...
+    min_x, min_y = np.min(pos_arr, axis=0)
+    max_x, max_y = np.max(pos_arr, axis=0)
+    scale_x = max_x - min_x
+    scale_y = max_y - min_y
+    pos_arr = pos_arr - np.array([min_x , max_y]) # shift to make (0, 0) the top left corner
+    pos_arr = pos_arr * np.array([1/scale_x, 1/scale_y]) # scale to make (1, -1) the bottom right corner
+    paddims = dim - 2
+    zeros = np.zeros((pos_arr.shape[0], paddims))
+    pos_arr = np.hstack((pos_arr*np.array([scale, scale]), zeros)) + center 
+    pos = {v: pos_arr[i] for i, v in enumerate(G.nodes())}
```
randomir, line 98: Would it make sense to de-dup this code by moving it into an utility function?

dwave_networkx/drawing/pegasus_layout.py
```diff
@@ -69,6 +69,7 @@ def pegasus_layout(G, scale=1., center=None, dim=2, crosses=False):
     >>> pos = dnx.pegasus_layout(G)
 
     """
+    import numpy as np
```
randomir, line 72: Unless there are performance concerns, or circular dependencies, imports should go to the top.

mahdiehmalekian: Done in commit `2d2485b`

dwave_networkx/drawing/zephyr_layout.py
mahdiehmalekian: Do you mean to use it in both `pegasus_layout.py` and `zephyr_layout.py`? It's unlikely it gets repeated elsewhere. But will do if you see fit for these two files. Your call.

randomir: Since you don't see this code used elsewhere, no need to extract it into a utility function. That would be a premature optimization.

Issue #234, opened 2023-12-04
Add graph relabeling and sublattice mappings to the docs


arcondello:
The methods added in https://github.com/dwavesystems/dwave-networkx/pull/211 are not currently included in the docs but should be.

qwriter:
I think we can close this issue.

PR #237, opened 2024-05-25
Fixed docstrings for graph relabeling and sublattice mapping functionality missing from reference documentation.

qwriter:
Also, edited those docstrings.

This is the fix for issue #234.

### DIFF
--- a/dwave_networkx/generators/chimera.py
+++ b/dwave_networkx/generators/chimera.py
@@ -252,7 +252,7 @@ def chimera_graph(m, n=None, t=None, create_using=None, node_list=None, edge_lis
 
 
 def find_chimera_indices(G):
-    """Attempts to determine the Chimera indices of the nodes in graph G.
+    """Attempt to determine the Chimera indices of the nodes in graph G.
 
     See the :func:`~chimera_graph()` function for a definition of a Chimera graph and Chimera
     indices.
@@ -496,7 +496,7 @@ class chimera_coordinates(object):
         return self._pair_repack(self.iter_linear_to_chimera, plist)
 
     def graph_to_linear(self, g):
-        """Return a copy of the graph g relabeled to have linear indices"""
+        """Return a copy of the graph `g` relabeled to have linear indices."""
         labels = g.graph.get('labels')
         if labels == 'int':
             return g.copy()
@@ -515,7 +515,7 @@ class chimera_coordinates(object):
             )
 
     def graph_to_chimera(self, g):
-        """Return a copy of the graph g relabeled to have chimera coordinates"""
+        """Return a copy of the graph `g` relabeled to have Chimera coordinates."""
         labels = g.graph.get('labels')
         if labels == 'int':
             return chimera_graph(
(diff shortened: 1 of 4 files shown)


dwave_networkx/generators/chimera.py
```diff
@@ -545,7 +545,7 @@ def __missing__(self, key):
 _chimera_coordinates_cache = __chimera_coordinates_cache_dict()
 
 def linear_to_chimera(r, m, n=None, t=None):
```
JoelPasvolsky, line 547: Is this function not documented anywhere?

dwave_networkx/generators/chimera.py
```diff
@@ -653,53 +653,55 @@ def mapping(q):
...
-    integer or coordinate labels.
-    
-    Academic note: the full group of isomorphisms of a Chimera graph includes 
-    mappings which permute tile indices on a per-row and per-column basis, in
-    addition to reflections and rotations of the grid of unit cells where 
-    rotations by 90 and 270 degrees induce a change in orientation.  The full
-    set of sublattice mappings would take those isomorphisms into account; we do
-    not undertake that complexity here.
-    
+    """Yield mappings from a Chimera graph into a larger Chimera graph.
```
JoelPasvolsky, line 656: Same comment about internal consistency

dwave_networkx/generators/chimera.py
```diff
@@ -653,53 +653,55 @@ def mapping(q):
...
+
     Parameters
     ----------
         source : NetworkX Graph
-            The Chimera graph that nodes are input from
+            The Chimera graph that nodes are input from.
         target : NetworkX Graph
-            The Chimera graph that nodes are input from
+            The Chimera graph that nodes are output to.
```
JoelPasvolsky, line 663: Nice catch!

dwave_networkx/generators/chimera.py
```diff
@@ -653,53 +653,55 @@ def mapping(q):
...
-            A function from nodes of the source graph, to nodes of the target
-            graph.  The offset used to generate this mapping is stored in
-            ``mapping.offset`` -- these can be collected and passed into 
-            ``offset_list`` in a later session.
+            A function from the nodes of the source graph to the nodes 
+            of the target graph.  The offset used to generate this mapping 
+            is stored in ``mapping.offset``, which can be collected and passed 
+            into ``offset_list`` in a later session.
+
+    Notes
```
JoelPasvolsky, line 677: I like the idea of using a different section to structure long descriptive content but these docstrings are now inconsistent with the rest of the documentation: the majority of docstrings use the Notes section for minor comments; see for example [maximum_independent_set](https://docs.ocean.dwavesys.com/en/stable/docs_dnx/reference/algorithms/generated/dwave_networkx.maximum_independent_set.html)

JoelPasvolsky (changes requested):
A couple of general comments: 

1. Please make all PRs on a new branch, not on ``main``
2. I made some comments on the first docstrings that apply to the rest as well, so I did not repeat them for all: these are about the use of the descriptive mode versus command for the first line, the use of back ticks for parameter references, and the Notes sections.

dwave_networkx/generators/chimera.py
qwriter: Could you elaborate? `linear_to_chimera` does have a docstring.

JoelPasvolsky: There's a ``linear_to_chimera``  method of ``chimera_coordinates`` defined on line 406 and a function ``linear_to_chimera`` defined on line 547, but I see only the first documented on line 40 of ``utilities.rst``

arcondello:
Closing as redundant to https://github.com/dwavesystems/dwave-networkx/pull/238

PR #240, opened 2024-06-19
Use default resource class for macos builds on CircleCi

randomir:
Support for `macos.x86.medium.gen2` is soon to be dropped.

See: https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,9 +62,6 @@ jobs:
     macos:
       xcode: "14.3.0"
 
-    environment:
-      HOMEBREW_NO_AUTO_UPDATE: 1
-
     steps:
       - checkout
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,22 +56,25 @@ jobs:
     parameters:
       python-version:
         type: string
+      xcode:
+        type: string
+        default: "14.3.0"
 
     resource_class: macos.m1.medium.gen1
 
     macos:
-      xcode: "14.3.0"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -56,34 +56,36 @@ jobs:
...
           name: Install python
           command: |
             pyenv install << parameters.python-version>> -s
 
       - save_cache: &save-cache-pyenv
-          key: v1-pyenv-{{ .Environment.CIRCLE_JOB }}-xcode-14.3.0
+          key: *brew-pyenv-cache-key
           paths:
+            - /Users/distiller/Library/Caches/Homebrew
+            - /usr/local/Homebrew
```
thisac, line 88: Why are these paths suddenly required? I'm assuming to find brew, but why not before?

thisac (approved):
LGTM

.circleci/config.yml
randomir: This is to cache homebrew cache, so `brew update` is quicker. You'll notice before this PR we only cached installed pythons in `.pyenv`. Now we cache brew, and **allow brew to update on each run** (to get the latest python provided by the latest pyenv) -- and all this with a 2x speed-up relative to previous setup -- macos jobs now take 2-3m vs 4-6m. Part of this speed-up is due to switch to M1, but we now also update brew cache on each run.

Anyway, this is the setup I use on several Ocean repos already.

PR #243, opened 2024-10-08
Feature/coloring: Four colorings for Zephyr and Pegasus, and a two coloring for Chimera

jackraymond:
Colorings of minimal chromatic number on Chimera, Pegasus and Zephyr graphs in the form of a map using the standard coordinate schemes.

### DIFF
--- a/dwave_networkx/generators/chimera.py
+++ b/dwave_networkx/generators/chimera.py
@@ -35,7 +35,7 @@ __all__ = ['chimera_graph',
            'linear_to_chimera',
            'chimera_sublattice_mappings',
            'chimera_torus',
-           'chimera_four_color',
+           'chimera_two_color',
            ]
 
 
@@ -869,7 +869,7 @@ def chimera_torus(m, n=None, t=None, node_list=None, edge_list=None):
     return G
 
 
-def chimera_four_color(q):
+def chimera_two_color(q):
     """
     Node color assignment sufficient for four coloring of a Chimera graph. 
 
@@ -886,11 +886,11 @@ def chimera_four_color(q):
     Examples
     ========
     A mapping of every qubit (default integer labels) in the Chimera[m, t]
-    graph to one of 4 colors
+    graph to one of 2 colors
     >>> m = 2
     >>> G = dnx.chimera_graph(m)
-    >>> colors = {q: dnx.chimera_four_color(dnx.chimera_coordinates(m,t).linear_to_chimera(q)) for q in G.nodes()}    # doctest: +SKIP
+    >>> colors = {q: dnx.chimera_two_color(dnx.chimera_coordinates(m,t).linear_to_chimera(q)) for q in G.nodes()}    # doctest: +SKIP
     
     """
     i, j, u, _ = q
-    return 2 * u + (i ^ j) & 1
+    return (i ^ j ^ u) & 1
--- a/dwave_networkx/generators/chimera.py
+++ b/dwave_networkx/generators/chimera.py
@@ -887,8 +887,8 @@ def chimera_two_color(q):
     A mapping of every qubit (default integer labels) in the Chimera[m, t]
     graph to one of 2 colors
     >>> m = 2
-    >>> G = dnx.chimera_graph(m)
-    >>> colors = {q: dnx.chimera_two_color(dnx.chimera_coordinates(m,t).linear_to_chimera(q)) for q in G.nodes()}    # doctest: +SKIP
+    >>> G = dnx.chimera_graph(m, coordinates=True)
+    >>> colors = {q: dnx.chimera_two_color(q) for q in G.nodes()}    # doctest: +SKIP
     
     """
     i, j, u, _ = q
--- a/dwave_networkx/generators/chimera.py
+++ b/dwave_networkx/generators/chimera.py
@@ -888,8 +888,7 @@ def chimera_two_color(q):
     graph to one of 2 colors
     >>> m = 2
     >>> G = dnx.chimera_graph(m, coordinates=True)
-    >>> colors = {q: dnx.chimera_two_color(q) for q in G.nodes()}    # doctest: +SKIP
-    
+    >>> colors = {q: dnx.chimera_two_color(q) for q in G.nodes()}
     """
     i, j, u, _ = q
     return (i ^ j ^ u) & 1
--- a/dwave_networkx/generators/chimera.py
+++ b/dwave_networkx/generators/chimera.py
@@ -35,6 +35,7 @@ __all__ = ['chimera_graph',
            'linear_to_chimera',
            'chimera_sublattice_mappings',
            'chimera_torus',
+           'chimera_four_color',
            ]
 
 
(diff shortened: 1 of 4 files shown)


kevinchern:
@arcondello  I review this

tests/test_regular_coloring.py
```diff
@@ -0,0 +1,69 @@
...
+            for topology_shape in topology_shapes:
+                G = graph(*topology_shape, coordinates=True)
+                col_dict = {q: color(q) for q in G.nodes}
+                self.assertTrue(np.all(np.unique(list(col_dict.values())) ==
+                                       np.arange(num_colors)))
+                self.assertTrue(all(col_dict[q1] != col_dict[q2]
+                                    for q1, q2 in G.edges),
+                                f'{topology_type}[{topology_shape}]')
+
+    def test_non_default_schemes(self):
```
kevinchern, line 52: remove this test; see above comment using `lambda`

tests/test_regular_coloring.py
```diff
@@ -0,0 +1,69 @@
...
+
+import dwave_networkx as dnx
+import numpy as np
+
+class TestRegularColoring(unittest.TestCase):
+
+    def test_valid(self):
+        test_cases = {'chimera': [(3,3), (4,4)],
+                      'pegasus': [(2,), (4,)],
+                      'zephyr': [(1,2), (3,4)]}
```
kevinchern, line 25: add tests for both zephyr schemes by, for example
```
...

'zephyr-scheme-0': [...],
'zephyr-scheme-1': [...],

...

color = lambda q: dnx.zephyr_color(q, scheme=1)
```

kevinchern (changes requested):

dwave_networkx/generators/pegasus.py
```diff
@@ -1303,3 +1304,28 @@ def relabel(u,w,k,z):
...
+    -------
+        color : int
+            Colors 0, 1, 2 or 3
+    Examples
+    ========
+    A mapping of every qubit (default integer labels) in the Pegasus[m]
+    graph to one of 4 colors
+    >>> m = 2
+    >>> G = dnx.pegasus_graph(m)
+    >>> colors = {q: dnx.pegasus_four_color(dnx.pegasus_coordinates(m).linear_to_zephyr(q)) for q in G.nodes()}    # doctest: +SKIP
```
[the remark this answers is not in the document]
jackraymond: Per offline suggestion, I added coordinates=True to simplify.

tests/test_regular_coloring.py
```diff
@@ -0,0 +1,59 @@
...
+                graph = dnx.chimera_graph
+                color = dnx.chimera_two_color
+                num_colors = 2
+            else:
+                raise ValueError('unrecognized topology')
+            
+            for topology_shape in topology_shapes:
+                G = graph(*topology_shape, coordinates=True)
+                col_dict = {q: color(q) for q in G.nodes}
+                self.assertTrue(np.all(np.unique(list(col_dict.values())) ==
```
kevinchern, line 51:
```suggestion
                self.assertSetEqual(set(col_dict.values()), set(range(num_colors)))
```

tests/test_regular_coloring.py
```diff
@@ -0,0 +1,59 @@
...
+                                       np.arange(num_colors)))
```
kevinchern, line 52:
```suggestion
```

kevinchern (approved):
can use set equality instead otherwise LGTM. passing the ball to poseidon/neptune (@arcondello / @randomir)

tests/test_regular_coloring.py
kevinchern: might go past the pep8 linewidth limit

randomir: nit: `assertEqual` already uses `assertSetEqual` by default if sets are compared, so no need to be specific.

nit 2: it's enough for one of the values to be set-like (have `.difference()` method), so no need to actually convert both to a set.

tests/test_regular_coloring.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2024 D-Wave Systems Inc.
```
randomir, line 1: New (c) should be used in any new code:
```suggestion
# Copyright 2024 D-Wave Inc.
```

tests/test_regular_coloring.py
```diff
@@ -0,0 +1,59 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import unittest
+
+import dwave_networkx as dnx
+import numpy as np
```
randomir, line 18: nit: [pep8 order](https://peps.python.org/pep-0008/#imports) is stdlib, 3rd-party, local

dwave_networkx/generators/chimera.py
```diff
@@ -866,3 +867,29 @@ def chimera_torus(m, n=None, t=None, node_list=None, edge_list=None):
...
+        color : int
+            Colors 0, 1, 2 or 3
+
+    Examples
+    ========
+    A mapping of every qubit (default integer labels) in the Chimera[m, t]
+    graph to one of 2 colors
+    >>> m = 2
+    >>> G = dnx.chimera_graph(m, coordinates=True)
+    >>> colors = {q: dnx.chimera_two_color(q) for q in G.nodes()}    # doctest: +SKIP
```
randomir, line 891: Why skip? (here and below)

tests/test_regular_coloring.py
```diff
@@ -0,0 +1,59 @@
...
+
+class TestRegularColoring(unittest.TestCase):
+
+    def test_valid(self):
+        test_cases = {'chimera': [(3,3), (4,4)],
+                      'pegasus': [(2,), (4,)],
+                      'zephyr-0': [(1,2), (3,4)],
+                      'zephyr-1': [(1,3), (3,2)]}
+
+        for topology_type, topology_shapes in test_cases.items():
```
randomir, line 28: Good candidate for `parameterized`, if you feel like simplifying. For an example, you can check [here](https://github.com/dwavesystems/dwave-cloud-client/blob/b2432fe441d304162ee2f3bac6249241366ece91/tests/auth/test_server.py#L32-L42).

randomir (approved):
LGTM.

Happy to merge, or can wait if you want to address a few minor comments I made below.

dwave_networkx/generators/chimera.py
jackraymond: Lazy habit

tests/test_regular_coloring.py
jackraymond: Next pull request

tests/test_regular_coloring.py
jackraymond: range() has no .difference() method, so is required this case.

randomir: Right, types have to match as well.

Issue #210, opened 2021-11-03
Enumerate known QPU families/topologies


randomir:
See https://github.com/dwavesystems/minorminer/pull/190#discussion_r741948893.

PR #248, opened 2024-12-25
add topology enums and dispatch decorator

boothby:
This is a preliminary patch to address #210 -- we add a `topology` enum to `dnx.topology`, and a decorator `topology_dispatch`. 

The topology enum is pretty simple, except that I want the machinery of py3.11's StrEnum (that is, `CHIMERA == 'chimera'`), so I fair-use'd barebones implementations of StrEnum and global_enum from the CPython3.11 library.

The `topology_dispatch` decorator can be used to define a generic function in one file, and then in other files, define specializations based on the topology family.  For example, I've added a `defect_free` generic function in `dnx.generators.common` and then added specializations for each of `CHIMERA`, `PEGASUS`, `ZEPHYR`.  Thus, to get a defect-free version of any qubit topology, one can simply call `dnx.generators.common.defect_free(G)`.

```python
#############
# common.py
#
@topology_dispatch
def defect_free(G):
    """Construct a defect-free topology based on the properties of G."""
    raise NotImplementedError(f"no defect-free generator defined for {G.graph.get('family')} graphs")


#############
# chimera.py
#
from ..topology import CHIMERA
from .common import defect_free

@defect_free.install_dispatch(CHIMERA)
def defect_free_chimera(G):
    """Construct a defect-free Chimera graph based on the properties of G."""
    attrib = G.graph
    family = attrib.get('family')
    if family != CHIMERA:
        raise ValueError("G must be constructed by dwave_networkx.chimera_graph")
    args = attrib['rows'], attrib['columns'], attrib['tile']
    kwargs = {'coordinates': attrib['labels'] == 'coordinate'}
    return chimera_graph(*args, **kwargs)
```

### DIFF
--- a/dwave_networkx/drawing/chimera_layout.py
+++ b/dwave_networkx/drawing/chimera_layout.py
@@ -20,13 +20,13 @@ Tools to visualize :term:`Chimera` lattices and weighted :term:`graph` problems 
 import networkx as nx
 from networkx import draw
 
-from dwave_networkx.drawing.qubit_layout import draw_qubit_graph, draw_embedding, draw_yield
-from dwave_networkx.generators.chimera import chimera_graph, find_chimera_indices, chimera_coordinates
-
+from dwave_networkx.drawing.qubit_layout import draw_qubit_graph, draw_embedding, draw_yield, qubit_layout
+from dwave_networkx.generators.chimera import chimera_graph, find_chimera_indices, chimera_coordinates, defect_free_chimera
+from ..topology import CHIMERA
 
 __all__ = ['chimera_layout', 'draw_chimera', 'draw_chimera_embedding', 'draw_chimera_yield']
 
-
+@qubit_layout.install_dispatch(CHIMERA, pop_kwargs=('scale', 'center', 'dim'))
 def chimera_layout(G, scale=1., center=None, dim=2):
     """Positions the nodes of graph ``G`` in a Chimera layout.
 
@@ -312,16 +312,5 @@ def draw_chimera_yield(G, **kwargs):
        the :func:`~networkx.drawing.nx_pylab.draw_networkx` ``node_color``
        or ``edge_color`` parameters are ignored.
     """
-    try:
-        assert(G.graph["family"] == "chimera")
-        m = G.graph["rows"]
-        n = G.graph["columns"]
-        t = G.graph["tile"]
-        coordinates = G.graph["labels"] == "coordinate"
-    except:
-        raise ValueError("Target chimera graph needs to have columns, rows, \
-        tile, and label attributes to be able to identify faulty qubits.")
-
-    perfect_graph = chimera_graph(m,n,t, coordinates=coordinates)
-
+    perfect_graph = defect_free_chimera(G)
     draw_yield(G, chimera_layout(perfect_graph), perfect_graph, **kwargs)
--- a/dwave_networkx/generators/chimera.py
+++ b/dwave_networkx/generators/chimera.py
@@ -26,7 +26,8 @@ from dwave_networkx.exceptions import DWaveNetworkXException
 
 from itertools import product
 
-from .common import _add_compatible_nodes, _add_compatible_edges, _add_compatible_terms
+from .common import _add_compatible_nodes, _add_compatible_edges, _add_compatible_terms, defect_free
+from ..topology import CHIMERA
 
 __all__ = ['chimera_graph',
            'chimera_coordinates',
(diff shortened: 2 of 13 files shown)


boothby:
After some thought, I wasn't a fan of my first approach; it was a bit too kludgy.  Re-reading the comments of dwavesystems/minorminer#210, Pau had given me a better suggestion for notation which I hadn't followed in the above.  My revised approach collects generic methods into the enumeration members.  Here, I avoid circular imports through the use of an `ImplementationHook` which is used to monkeypatch itself away during the initial import.  

With this new approach, I don't need to add generic functions to multiple files; instead, I collect their names into `topology.py`.  Instead of that `defect_free` function, I instead make a `defect_free_graph` slot for the `Topology` enum members, and when I implement `defect_free_chimera`, it looks like:

```python
@CHIMERA.defect_free_graph.implementation
def defect_free_chimera(G):
    ...
```
After making this change, my choice to use `StrEnum` had a negative impact: the namespace was polluted with the various methods of `str`.  On reflection, I think that it's important to preserve a few features of `str` since I'm populating the `dwave_networkx` graph `"family"` property with these enum members -- namely, `__eq__`, `__hash__`, and `__str__` should be enough for backwards compatibility and future convenience.

dwave_networkx/topology.py
```diff
@@ -0,0 +1,77 @@
+from enum import Enum, auto
+
+try:
+    from enum import global_enum
+except ImportError:
+    # delete after Py3.10 EOL (2026/10)
+    import sys
```
arcondello, line 7: For a slightly cleaner namespace, I'd move this inside the function body.

dwave_networkx/topology.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def global_enum(cls):
```
arcondello, line 9: Might also be worth a comment that this implementation is backported from the Python enum library

tests/common.py
```diff
@@ -0,0 +1,2 @@
+def graphs_equal(G, H):
```
arcondello, line 1: IMO it would be less confusing to a reader to do
```
self.assertIs(type(G), type(H))
self.assertTrue(nx.graphs_equal(G, H))
```
in each of the places we use this method. For users coming from `NetworkX` they'll likely to not immediately notice that this is a different method. IMO the slightly higher verbosity is worth it.

If we do keep this, needs an apache header.

arcondello (commented):
Did a pass for code. Needs to ruminate on the design a bit. Will try to do another pass later today or tomorrow.

tests/common.py
boothby: I wrote this abomination because we're still testing networkx2.4, which doesn't have a graphs_equal method :upside_down_face:

arcondello: In that case I'd rather raise the floor. High time we require 3.0

boothby: But you give me a better idea here: to subclass UnitTest and make each equality check an assertEqual.  Then we'll have a strictly better `self.assertGraphsEqual` method to use throughout the library.

arcondello: IMO more explicit would be a mixin rather than a intermediate class.
```
class GraphMixin:
    def assertGraphsEqual(...): ...
```

arcondello: I guess an intermediate class would allow you to just use `self.assertIs(...)` etc without messing about with abstract methods. So maybe my first idea wasn't my best.

boothby: I proposed an intermediate class and started writing it and realized that a mixin totally works -- no need for abstract methods :)

dwave_networkx/generators/chimera.py
```diff
@@ -252,6 +254,18 @@ def checkadd(v, q):
     return G
 
 
+@CHIMERA.defect_free_graph.implementation
+def defect_free_chimera(G):
+    """Construct a defect-free Chimera graph based on the properties of G."""
+    attrib = G.graph
+    family = attrib.get('family')
+    if family != CHIMERA:
```
thisac, line 262:
```suggestion
    if family is not CHIMERA:
```
Similarly in other files.

dwave_networkx/generators/chimera.py
```diff
@@ -252,6 +254,18 @@ def checkadd(v, q):
     return G
 
 
+@CHIMERA.defect_free_graph.implementation
+def defect_free_chimera(G):
+    """Construct a defect-free Chimera graph based on the properties of G."""
```
thisac, line 259: Missing Parameters and Returns.

dwave_networkx/drawing/chimera_layout.py
```diff
@@ -71,7 +72,7 @@ def chimera_layout(G, scale=1., center=None, dim=2):
 
     # now we get chimera coordinates for the translation
     # first, check if we made it
-    if G.graph.get("family") == "chimera":
+    if G.graph.get("family") == CHIMERA:
```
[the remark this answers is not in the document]
thisac: I'm not particularly concerned, but I can't say that I'm aware of how much (if at all) this might disrupt some users that have old pickled instances of graphs. If so, keeping it as an eq comparison and adding a comment about it being kept for backward compatibility (potentially a deprecation note for the future).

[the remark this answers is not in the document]
thisac: These changes should preferably be tracked in a changelog, possibly under breaking.

PR #267, opened 2026-04-08
Remove deprecated license classifier

thisac:
Setuptools>=77 now raises
```
setuptools.errors.InvalidConfigError: License classifiers have been superseded by license expressions (see https://peps.python.org/pep-0639/). Please remove:

      License :: OSI Approved :: Apache Software License
```
since its deprecation. See [PEP 639](https://peps.python.org/pep-0639) and [#4833](https://github.com/pypa/setuptools/pull/4833) for details.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ readme = {file = "README.rst", content-type = "text/x-rst"}
 license = "Apache-2.0"
 license-files = ["LICEN[CS]E*"]
 classifiers = [
-    "License :: OSI Approved :: Apache Software License",
     "Operating System :: OS Independent",
     "Programming Language :: Python :: 3 :: Only",
     "Programming Language :: Python :: 3.10",


randomir (approved):
LGTM!

But the new setuptools behavior screams bug to me! Zero grace period / *deprecation period* -- but immediate build error! The PEP says the tools MAY raise an error.. So the better approach would be they first raised a warning, then turned it into an error...

thisac:
> So the better approach would be they first raised a warning, then turned it into an error...

Completely agree with that! Also understood it as optional, but the error is there and this fix isn't really intrusive. :face_with_diagonal_mouth:

randomir:
> > So the better approach would be they first raised a warning, then turned it into an error...
> 
> Completely agree with that! Also understood it as optional, but the error is there and this fix isn't really intrusive. 🫤

Also, they only raise a warning under the same conditions if the config is in `setup.cfg`/`setup.py`:
```
SetuptoolsDeprecationWarning: License classifiers are deprecated.
```
and raise an error if `pyproject.toml` is used.

PR #276, opened 2026-07-16
Maximum bipartite matching

mahdiehmalekian:
Code for finding maximum bipartite matching via the Hopcroft-Karp algorithm.

AI use: I used AI to generate the code, checked the code myself.

### DIFF
--- a/dwave/graphs/algorithms/matching.py
+++ b/dwave/graphs/algorithms/matching.py
@@ -12,7 +12,8 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
-from collections.abc import Hashable
+from collections.abc import Hashable, Iterable, Mapping
+from collections import deque
 
 import dimod
 from dimod.decorators import graph_argument
@@ -20,6 +21,7 @@ from dimod.typing import GraphLike
 
 __all__ = ['maximal_matching',
            'min_maximal_matching',
+           'maximum_bipartite_matching',
            ]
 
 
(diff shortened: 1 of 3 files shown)


dwave/graphs/algorithms/matching.py
```diff
@@ -12,22 +12,26 @@
...
-def maximal_matching(graph: GraphLike,
-                     sampler: dimod.Sampler,
-                     **sampler_args,
-                     ) -> set[Hashable]:
+@graph_argument("graph")
+def maximal_matching(
+    graph: GraphLike,
+    sampler: dimod.Sampler,
+    **sampler_args,
+) -> set[Hashable]:
```
randomir: Please limit your changes in a PR to features/fixes actually added. Generally, avoid touching/reformatting lines/blocks of unrelated code to minimize diff and preserve git blame.

Also, adapt the style of code you add to the surrounding function, module, package (PEP-8).

Applies here and elsewhere.

dwave/graphs/algorithms/matching.py
```diff
@@ -127,3 +131,93 @@ def min_maximal_matching(graph: GraphLike,
 
     # the matching are the edges that are 1 in the sample
     return set(tuple(edge) for edge, val in sample.items() if val > 0)
+
+
+def maximum_bipartite_matching(
+    X: Iterable[Hashable],
+    Y: Iterable[Hashable],
+    edges: Mapping[Hashable, Iterable[Hashable]],
+) -> dict:
```
randomir:
```suggestion
) -> Mapping[Hashable, Hashable]:
```

dwave/graphs/algorithms/matching.py
```diff
@@ -127,3 +131,93 @@ def min_maximal_matching(graph: GraphLike,
...
+            ``Y``-neighbors. Vertices with no edges may be omitted.
+
+    Returns:
+        A dict mapping each matched ``x`` to its matched ``y``. Unmatched
+        vertices are absent, so ``len(result)`` is the size of the maximum
+        matching. If several maximum matchings exist, one of them is
+        returned; which one depends on iteration order.
+
+    Raises:
+        KeyError: If ``edges`` references a ``y`` that is not in ``Y``.
```
randomir: What if it references `x` not in `X`?

dwave/graphs/algorithms/matching.py
```diff
@@ -127,3 +131,93 @@ def min_maximal_matching(graph: GraphLike,
...
+
+    Note:
+        The augmenting-path search (``dfs``) is recursive with depth bounded
+        by the length of the augmenting path, so extremely large or
+        pathological graphs (deep augmenting paths) could approach Python's
+        recursion limit. For typical matching problem sizes this isn't a
+        concern.
+    """
+    X = list(X)
```
randomir:
```suggestion
```
Is this cast really needed? Looks like the only requirement for `X` in the code below is to be iterable.

dwave/graphs/algorithms/matching.py
```diff
@@ -127,3 +131,93 @@ def min_maximal_matching(graph: GraphLike,
...
+    Raises:
+        KeyError: If ``edges`` references a ``y`` that is not in ``Y``.
+
+    Note:
+        The augmenting-path search (``dfs``) is recursive with depth bounded
+        by the length of the augmenting path, so extremely large or
+        pathological graphs (deep augmenting paths) could approach Python's
+        recursion limit. For typical matching problem sizes this isn't a
+        concern.
+    """
```
randomir, line 180: Worth mentioning Hopcroft-Karp algorithm is used. Also, provide a ref, e.g. Wikipedia.

randomir (commented):
The implementation looks correct (when compared to the pseudo-code on Wikipedia), apart from a couple of pruning condition checks missing in your implementation, in the BFS and the DFS functions -- resulting in a greater runtime complexity in the worst case.

But my main question is what's the benefit of your implementation over the existing one in NetworkX, [`hopcroft_karp_matching()`](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.bipartite.matching.hopcroft_karp_matching.html)?

dwave/graphs/algorithms/matching.py
mahdiehmalekian: Yes, not needed.

dwave/graphs/algorithms/matching.py
mahdiehmalekian: Reverted the formatting changes in this file and in the test file. [dbda40f](https://github.com/dwavesystems/dwave-graphs/pull/276/commits/dbda40f634f0abd2f3ead25f14646cf03e4c9f08)

dwave/graphs/algorithms/matching.py
mahdiehmalekian: Removed this and made it explicit that ``edges`` must have ``x`` in ``X`` and ``y`` in ``Y``. [dbda40f](https://github.com/dwavesystems/dwave-graphs/pull/276/commits/dbda40f634f0abd2f3ead25f14646cf03e4c9f08)

dwave/graphs/algorithms/matching.py
mahdiehmalekian: Done in [dbda40f](https://github.com/dwavesystems/dwave-graphs/pull/276/commits/dbda40f634f0abd2f3ead25f14646cf03e4c9f08)

mahdiehmalekian:
Re. the missing pruning step, thank you for flagging it. I added it in [dbda40f](https://github.com/dwavesystems/dwave-graphs/pull/276/commits/dbda40f634f0abd2f3ead25f14646cf03e4c9f08)

Re. your main question, great point.
This is indeed the same algorithm as `networkx.bipartite.hopcroft_karp_matching` (except for the greedy warm-start and the BFS shortest-path prune added in [dbda40f](https://github.com/dwavesystems/dwave-graphs/pull/276/commits/dbda40f634f0abd2f3ead25f14646cf03e4c9f08) in this implementation). However, the main advantage of this implementation is for usage in the clique embedder in `minorminer` where this function is called in a hot loop where the inputs are already in the exact form this function needs. The two vertex sets are known by construction (vertical vs horizontal runs in the clique embedder), and the edges are already a plain adjacency dict. Using `networkx.bipartite.hopcroft_karp_matching` instead would require constructing a `nx.Graph` (node/edge insertion, attribute dicts) on every call, passing the bipartition we already know as `top_nodes`, and then inverting the symmetric result back into the one-directional {vertical run: horizontal run} map our caller expects. This custom version skips all of that and returns the exact contract the caller consumes.

dwave/graphs/algorithms/matching.py
```diff
@@ -127,3 +129,96 @@ def min_maximal_matching(graph: GraphLike,
...
+Notes:
+        Uses the Hopcroft-Karp algorithm (O(sqrt(|V|) . |E|)), where 
+        |V| = |X| + |Y| is the number of vertices and |E| is the total
+        number of edges. See
+        <https://en.wikipedia.org/wiki/Hopcroft%E2%80%93Karp_algorithm>
+
+        The augmenting-path search (``dfs``) is recursive with depth bounded
+        by the length of the augmenting path, so extremely large or
+        pathological graphs (deep augmenting paths) could approach Python's
+        recursion limit. For typical matching problem sizes this isn't a
```
randomir: Could you please elaborate what does "typical" mean here?

dwave/graphs/algorithms/matching.py
```diff
@@ -12,18 +12,21 @@
...
 from dimod.typing import GraphLike
 
-__all__ = ['maximal_matching',
-           'min_maximal_matching',
-           ]
+__all__ = [
+    "maximal_matching",
+    "min_maximal_matching",
+    "maximum_bipartite_matching",
+]
```
randomir: Fix indent.

dwave/graphs/algorithms/matching.py
```diff
@@ -12,18 +12,21 @@
...
 
 
-@graph_argument('graph')
+@graph_argument("graph")
```
randomir: Revert
```suggestion
@graph_argument('graph')
```

dwave/graphs/algorithms/matching.py
```diff
@@ -73,7 +75,7 @@ def maximal_matching(graph: GraphLike,
     return set(tuple(edge) for edge, val in sample.items() if val > 0)
 
 
-@graph_argument('graph')
+@graph_argument("graph")
```
randomir:
```suggestion
@graph_argument('graph')
```

dwave/graphs/algorithms/matching.py
```diff
@@ -127,3 +129,96 @@ def min_maximal_matching(graph: GraphLike,
...
+            ``edges`` must map vertices in ``X`` to vertices in ``Y``; passing
+            keys or neighbors outside these sets is undefined.
+
+    Returns:
+        A dict mapping each matched ``x`` to its matched ``y``. Unmatched
+        vertices are absent, so ``len(result)`` is the size of the maximum
+        matching. If several maximum matchings exist, one of them is
+        returned; which one depends on iteration order.
+
+Notes:
```
randomir: Indent.
```suggestion
    Notes:
```

dwave/graphs/algorithms/matching.py
```diff
@@ -127,3 +129,96 @@ def min_maximal_matching(graph: GraphLike,
...
+        dist[None] = INF  # distance to the nearest free y (shortest aug. path)
+        while q:
+            x = q.popleft()
+            if dist[x] < dist[None]:  # prune: don't expand past the shortest layer
+                for y in edges.get(x, ()):
+                    matched_x = pair_y[y]  # None if y is free
+                    if dist[matched_x] == INF:  # free y, or x not yet layered
+                        dist[matched_x] = dist[x] + 1
+                        q.append(matched_x)
+        return dist[None] is not INF
```
randomir: `float('inf')` is not a singleton, best not to compare it with the `is` operator.
```suggestion
        return dist[None] != INF
```

For example:
```python
>>> (1e308 * 2) is float('inf')
False
>>> (1e308 * 2) == float('inf')
True
```

randomir:
> Re. your main question, great point. This is indeed the same algorithm as `networkx.bipartite.hopcroft_karp_matching` (except for the greedy warm-start and the BFS shortest-path prune added in [dbda40f](https://github.com/dwavesystems/dwave-graphs/pull/276/commits/dbda40f634f0abd2f3ead25f14646cf03e4c9f08) in this implementation). However, the main advantage of this implementation is for usage in the clique embedder in `minorminer` where this function is called in a hot loop where the inputs are already in the exact form this function needs. The two vertex sets are known by construction (vertical vs horizontal runs in the clique embedder), and the edges are already a plain adjacency dict. Using `networkx.bipartite.hopcroft_karp_matching` instead would require constructing a `nx.Graph` (node/edge insertion, attribute dicts) on every call, passing the bipartition we already know as `top_nodes`, and then inverting the symmetric result back into the one-directional {vertical run: horizontal run} map our caller expects. This custom version skips all of that and returns the exact contract the caller consumes.

Perhaps worth mentioning that in the docs?

dwave/graphs/algorithms/matching.py
mahdiehmalekian: Done in [e693d74](https://github.com/dwavesystems/dwave-graphs/pull/276/commits/e693d74bb6422096ca873fd01de8bea8c99eb72d)

dwave/graphs/algorithms/matching.py
mahdiehmalekian: Done in [afeb155](https://github.com/dwavesystems/dwave-graphs/pull/276/commits/afeb15555dd7cf40084fb2eb9ec341b9da1a28d8)

mahdiehmalekian:
Reverted formatting where applicable and added a note comparing with `networkx.bipartite.hopcroft_karp_matching` .

randomir:
Thank you @mahdiehmalekian for addressing all the review comments.

And just to build a strong(er) case for code duplication (a code that's already tested and maintained by the networkx team) -- could you please run a quick benchmark against the nx implementation in the clique embedder hot loop -- just to get a sense if this code provides a 10% or a 10x improvement?

Also, could you please a [release note](https://github.com/dwavesystems/dwave-graphs#release-notes)?

mahdiehmalekian:
I benchmarked this custom version against a `networkx` wrapper that builds the `nx.Graph`, calls `hopcroft_karp_matching(G, top_nodes=X)`, then invert the result back into the one-directional `{v-run: h-run}` map our caller in clique embedder expects. Inputs were captured from a real run on an Adv2 size graph, Z(m =12, t=4,) 5% node / 1% edge faults; 800 calls.

 **~5.6x faster in aggregate** over the real call; **median 6x per call**. So it's much closer to the 10x end than 10%.

mahdiehmalekian:
Making the PR to `minorminer/utils/_clique` instead per Slack conversation

PR #277, opened 2026-08-27
Migrate automorphism utilities from dwave-experimental

SebastianGitt:
## Summary

Migrate the automorphism algorithms from dwave-experimental to dwave-graphs.

- Move `automorphism_generation.py` to `dwave/graphs/algorithms/automorphism.py`
- Migrate the associated `test_automorphism.py` tests
- Update imports and expose the automorphism functionality through the `dwave.graphs` namespace
- Preserve the implementation and test history from `dwave-experimental`


dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1171 @@
...
+                break
+
+        compare_result = 2
+        if self._best_perm_exist:  # if a leaf node has been reached previously
+
+            if self._compare_adj:
+                perm_candidate = list(chain.from_iterable(p for p in partition if p is not None))
+                compare_result = self._compare(perm_candidate, first_split)
+            else:
+                compare_result = trace.tobytes() == self._trace_history[self._depth - 1]
```
randomir, line 586: **[critical] Single-component path certifies leaves as automorphic by 1-WL trace equality alone, producing wrong automorphism groups for regular/WL-hard graphs.**

Reproducers (executed against this branch):
- `schreier_rep(nx.frucht_graph()).num_automorphisms` returns **479001600** (= 12!) for a graph with trivial automorphism group (|Aut| = 1); all 66 generators stored in `u_vector` are verified non-automorphisms, so `vertex_orbits`, `edge_orbits`, and `sample_automorphisms` are garbage too.
- Shrikhande graph: returns **9153720576000** instead of 192.
- ~54% of random regular graphs tested (193/359) miscount.

The correct adjacency comparison (`_compare_adj`) already exists but is only enabled on the multi-component path (which tests correct on the same graphs). The docstring's "identical pruning capability for graphs of interest" caveat from dwave-experimental does not hold for the arbitrary `nx.Graph` inputs this 1.0 public API now accepts. Consider enabling adjacency comparison at leaves (or verifying candidate automorphisms before `_enter`) on the single-component path as well.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1171 @@
...
+        >>> u_vector = [
+        ...     [np.array([0, 1, 4, 3, 2, 6, 5, 7])],
+        ...     [np.array([2, 1, 4, 3, 0, 7, 5, 6]), np.array([4, 1, 0, 3, 2, 6, 7, 5])],
+        ...     [np.array([0, 3, 2, 1, 4, 5, 6, 7])],
+        ... ]
+        >>> nodes = list(range(8))
+        >>> vertex_orbits(u_vector, nodes)
+        [[0, 2, 4], [1, 3], [5, 6, 7]]
+    """
+    if not u_vector:
```
randomir, line 809: **[bug] `vertex_orbits`/`edge_orbits` early-return for empty `u_vector` ignores `index_to_node`, so `*_original_labels` return internal indices for rigid graphs.**

For the asymmetric tree `nx.Graph([('a','b'),('b','c'),('c','d'),('d','e'),('e','f'),('c','g')])` (`u_vector == []`), `result.vertex_orbits_original_labels` returns `[[0], [1], ..., [6]]` instead of `[['a'], ..., ['g']]`, because this line returns `[[x] for x in nodes]` before the label mapping is applied. Same defect at line 868 for edges, which also skips tuple normalization. Verified by execution.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1171 @@
...
+        ...
+        >>> graph = nx.cycle_graph(8)
+        >>> result = schreier_rep(graph)
+        >>> sample_automorphisms(result.u_vector, seed=42)
+        [array([3, 4, 5, 6, 7, 0, 1, 2])]
+        >>> sample_automorphisms(result.u_vector, num_samples=2, seed=42)
+        [array([3, 4, 5, 6, 7, 0, 1, 2]), array([6, 5, 4, 3, 2, 1, 0, 7])]
+    """
+    rng = np.random.default_rng(seed)
+    num_nodes = len(u_vector[0][0])
```
randomir: **[bug] `sample_automorphisms` crashes with `IndexError` when `u_vector` is empty (graph with only the identity automorphism) instead of returning identity permutations.**

Reproducer: `graph` = smallest asymmetric tree (path 0-1-2-3-4-5 plus pendant 6 on node 2); `schreier_rep(graph).u_vector == []`; `sample_automorphisms(result.u_vector)` executes `num_nodes = len(u_vector[0][0])` and raises `IndexError: list index out of range`. Verified by execution. Note the identity fallback cannot even determine `num_nodes` from an empty `u_vector`, so the signature may need to accept it.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1171 @@
...
+
+    # group isomorphic components together
+    components = [ctx._graph.subgraph(c).copy() for c in nx.connected_components(ctx._graph)]
+
+    unique_components = {}
+    for comp in components:
+        # extract per-component coloring keyed by the component's node labels
+        comp_coloring = None
+        if graph_coloring is not None:
+            comp_coloring = {v: graph_coloring[index_to_node[v]] for v in comp.nodes()}
```
randomir: **[bug] Multi-component `schreier_rep` bypasses `graph_coloring` validation: missing nodes raise bare `KeyError` instead of the documented `ValueError`, and extra nodes are silently accepted.**

The global `SchreierContext` on this path is built *without* the coloring, so validation in `__init__` never runs. Verified by execution:
- `schreier_rep(nx.disjoint_union(nx.path_graph(2), nx.path_graph(2)), graph_coloring={0:0,1:0,2:0})` raises `KeyError: 3` from this dict comprehension, while the identical mistake on a connected graph raises `ValueError('graph_coloring is missing nodes: ...')`.
- `graph_coloring={0:0,1:0,2:0,3:0,99:5}` is silently accepted here but rejected on the single-component path.

randomir (commented):
Automated code review of PR #277 (automorphism migration). 10 verified findings posted as inline comments; the top 4 correctness findings were confirmed by executing reproducers against this branch.

One finding cannot be anchored inline because the file is not part of this diff:

**docs/algorithms.rst — new public API is undocumented.** The eight newly exported names (`schreier_rep`, `SchreierContext`, `sample_automorphisms`, `vertex_orbits`, `edge_orbits`, `array_to_cycle`, `mult`, `inv`) are absent from the Sphinx docs, unlike every other algorithms module which has an automodule/autosummary section in `docs/algorithms.rst`. The feature announced in the release note is invisible in the rendered 1.0 API reference.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1171 @@
...
+        else:
+            self._graph_coloring_internal = None
+
+        self._nodes: list[int] = list(graph.nodes())
+        self._num_nodes: int = graph.number_of_nodes()
+        self._graph_edges: list[tuple[int, int]] = list(graph.edges())
+        self._neighbours: list[set[int]] = [set(graph.neighbors(i)) for i in range(self._num_nodes)]
+        self._graph: nx.Graph = graph
+
+        self._num_samples: int | None = num_samples
```
randomir: **[api] Public parameters `num_samples` and `seed` are stored (`self._num_samples`, `self._rng = random.Random(seed)`) but never read anywhere — the documented behavior does not exist.**

The docstrings promise "Number of samples to use for generating new coset representatives" and "seed for reproducibility", but neither attribute is ever consumed, so callers trying to bound composition work or vary results by seed silently get identical behavior; the `import random` exists solely to feed the dead attribute. Either implement or remove the parameters before they ship as public API.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1171 @@
...
+        u_vector: Coset representatives grouped by stabilizer index.
+        nodes: List of vertex indices used to return trivial orbits when ``u_vector`` is empty.
+        index_to_node: An optional dictionary for returning orbits with their original node labels.
+
+    Returns:
+        A list of orbits, each orbit is a list of vertex indices.
+
+    Example:
+        >>> import numpy as np
+        >>> from dwave.experimental.automorphism import vertex_orbits
```
randomir: **[docs] All six docstring examples still import from the old module path `dwave.experimental.automorphism`** (lines 780, 850, 931, 971, 993, 1139) — the migration adapted the test imports but not the documented examples.

A dwave-graphs user copy-pasting `>>> from dwave.experimental.automorphism import vertex_orbits` gets `ModuleNotFoundError` (dwave-experimental is not a dependency), and `docs/conf.py` enables `sphinx.ext.doctest`, so once the module is added to the docs all six doctest blocks fail on the import line.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1171 @@
...
+                return
+
+        if first_split == self._num_nodes - 1:  # leaf node reached
+            self._leaf_nodes += 1
+
+            if not self._best_perm_exist:
+                self._best_perm_exist = True
+                self._best_perm[:] = list(chain.from_iterable(partition))
+
+            elif compare_result == 2:
```
randomir, line 598: **[maintainability] `compare_result` is int 0/1/2 in adjacency mode but `bool` in trace mode (line 583), so this `elif compare_result == 2` branch is unreachable in trace mode and `_best_perm` stays frozen at the first DFS leaf.**

On the single-component path the "canonical" `_best_perm` (and hence `_certificate()`, which `test_kreher` asserts on this path) is just the first leaf reached, never refined by greater-than comparisons — the certificate is not canonical there. Any future edit assuming the documented 0/1/2 contract (`True == 2` is `False`, `True == 1` is `True`) breaks silently in trace mode. Tightly coupled to the leaf-comparison finding at line 583: fixing that should also unify this dual typing.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1171 @@
...
+                    partition[num_colors[0]] = set()
+                    if in_refine_stack[color_to_split] or i != largest_subcell_degree:
+                        refine_stack.append(degree_to_new_color[i])
+                        in_refine_stack[degree_to_new_color[i]] = 1
+                    num_colors[0] += 1
+
+        for v in active_vertices[color_to_split]:
+            new_color = degree_to_new_color[color_degree[v]]
+            if new_color != color_to_split:
+                partition[color_to_split] = partition[color_to_split] - {v}  # must create new obj
```
randomir: **[perf] `_split_up_color` rebuilds the entire source cell set once per moved vertex (`partition[color_to_split] = partition[color_to_split] - {v}`), making each split O(cell_size^2) in the refinement hot path.**

Refinement runs at every search-tree node; the first split of an n-node graph moves O(n) vertices out of the initial n-vertex cell, copying the whole remaining set each time (~12.5M element copies at the root refine of a ~5000-node zephyr graph, repeated across thousands of nodes). The copy-on-write constraint ("must create new obj" — ancestor partition lists share the set) only requires one fresh object per split call: copy the cell once before the vertex loop, `discard()` in place, assign back once — O(cell_size) total.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1171 @@
...
+    if nx.number_connected_components(graph) == 1:
+        ctx = SchreierContext(
+            graph, num_samples=num_samples, seed=seed, graph_coloring=graph_coloring
+        )
+        initial_partition, trace, color, num_colors = ctx._initial_partition()
+
+        ctx._canon(initial_partition, trace, color, num_colors)
+        return ctx
+
+    # relabel vertices so components have contiguous labels
```
randomir: **[cleanup] The multi-component branch re-implements the sorted contiguous relabeling that `SchreierContext.__init__` already performs, then overwrites the context's private `_index_to_node`/`_node_to_index` from outside.**

Two copies of the label-mapping logic must be kept in sync, and the external mutation of private attributes bypasses the class invariant. The cost is already realized: the two paths have diverged on `graph_coloring` validation (line 1075 finding) and on leaf-comparison semantics (line 583 finding). Letting `SchreierContext` accept an optional precomputed node ordering (or extracting one relabel helper) removes the duplicate path.

dwave/graphs/algorithms/automorphism.py
SebastianGitt: Thanks, I was aware of this one already but doing a comparison at every leaf results in a dramatic decrease in performance and it only fails for a few niche graphs. I did document that the trace comparison at the leaves is only guaranteed to be correct for a subset of graphs. That said, I have been investigating a fix that does correctly handle random regular graphs and Frucht graphs (and I guess Shrikhande graphs too?) without a performance decrease. I was planning on adding it as a follow-up PR, but I can add it here instead.

dwave/graphs/algorithms/automorphism.py
SebastianGitt: In trace mode the certificate will not be canonical, I included this test just to make sure future changes didn't unknowingly change the certificate.

dwave/graphs/algorithms/automorphism.py
SebastianGitt: These are actually different. For the multi-component branch, contiguity is enforced on an intra-component level. But passing the node ordering into `SchreierContext` is a good idea regardless.

dwave/graphs/algorithms/automorphism.py
SebastianGitt: After doing some benchmarking it doesn't look like this makes any difference for the cases I tried, likely due to the fact that the majority of the cell splits don't involve very many nodes. I'll implement it regardless, as there may be certain graphs where it is noticeable.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1187 @@
...
+__all__ = [
+    "array_to_cycle",
+    "edge_orbits",
+    "inv",
+    "mult",
+    "sample_automorphisms",
+    "SchreierContext",
+    "schreier_rep",
+    "vertex_orbits",
+]
```
thisac, line 31: I'm not sure it makes sense moving all of these up into the `dwave.graphs` namespace. It seems to me better to keep some of these just accessible via `dwave.graph.algorithms.automorphism` or `dwave.graph.algorithms`. 🤔

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1187 @@
...
+            if sample_indices[i] >= 0:
+                g = u_i[sample_indices[i]]
+                g_product = mult(g, g_product)
+
+        sampled_automorphisms.append(g_product)
+
+    return sampled_automorphisms
+
+
+def mult(alpha: NDArray[np.intp], beta: NDArray[np.intp]) -> NDArray[np.intp]:
```
thisac: `mult()` and `inv()` seem a bit too general to be in this module. Perhaps either there's a better place for them, or hide them for internal module use only.

thisac: Also, is `mult()` even required, or could it just be in-lined instead?

releasenotes/notes/add-automorphism-888c471b46819be3.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    Add graph automorphism algorithms migrated from ``dwave-experimental``,
+    including functions for computing Schreier representations, sampling
+    automorphisms, and determining vertex and edge orbits.
```
thisac: I would probably skip mentioning that this is migrated from dwave-experimental, and just have this be mentioned as a new feature.

dwave/graphs/algorithms/automorphism.py
thisac: Although, currently, everything from `algorithms` is currently accessible via `dwave.graphs` @randomir

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1187 @@
...
+            j_canon = j_nodes[j_canon_perm]
+
+            u_global = np.arange(graph_nnodes)
+            u_global[i_canon], u_global[j_canon] = u_global[j_canon], u_global[i_canon]
+            ctx._enter(u_global, mode=EnterMode.RECURSE_ONCE)
+
+    return ctx
+
+
+def array_to_cycle(
```
thisac: This also seems more general than just for automorphism related functionality, right?

dwave/graphs/algorithms/automorphism.py
randomir: I agree we should scale it down a bit. Generally I think we're too keen on star imports and including symbols in `__all__` across Ocean. IMO, only the key ones should be propagated to the root. Modules like `dwave.graph.algorithms` should be imported at root, but not _everything_ from algorithms in the root.

E.g. this should work `dwave.graph.algorithms.SchreierContext`, but not `dwave.graphs.SchreierContext`.

dwave/graphs/algorithms/automorphism.py
SebastianGitt: Having `mult()` as its own function is useful for benchmarking purposes and makes things a bit more readable in my opinion. But I agree that making them for internal use only makes sense.

dwave/graphs/algorithms/automorphism.py
SebastianGitt: Yes this is just a general utility for converting permutations from one-line notation to cycle notation. I can remove it.

dwave/graphs/algorithms/automorphism.py
SebastianGitt: On second thought, I went with your suggestion of in-lining `mult()` (and also `inv()`) since they are only used in one or two places in the current code, and the solver doesn't spend nearly as much time in them as it did in previous versions of the algorithm, so it's not as important for benchmarking purposes anymore.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1085 @@
...
+
+
+class SchreierContext:
+    """This object holds mutable states used throughout the automorphism calculation.
+
+    Args:
+        graph: A NetworkX Graph object representing the input graph.
+        num_samples: Number of samples to use for generating new coset representatives
+            from the existing set. If not provided, all coset representatives are used.
+        seed: Seed used for reproducibility. Defaults to 42.
```
randomir: Stale docs:
```suggestion
```

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1085 @@
...
+
+    def __init__(
+        self,
+        graph: nx.Graph,
+        graph_coloring: Mapping[Hashable, int] | None = None,
+        node_order: list[Hashable] | None = None,
+    ) -> None:
+        if node_order is None:
+            node_order = sorted(graph.nodes())
+        elif set(node_order) != set(graph.nodes()):
```
randomir: Add a length check here to prevent a later failure in case of duplicates in `node_order`.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1085 @@
...
+    """
+
+    def __init__(
+        self,
+        graph: nx.Graph,
+        graph_coloring: Mapping[Hashable, int] | None = None,
+        node_order: list[Hashable] | None = None,
+    ) -> None:
+        if node_order is None:
+            node_order = sorted(graph.nodes())
```
randomir: nit: Sorting will fail for mixed types of node labels (like str, int), although these are legitimate nx graphs. Consider using `list` if `sorted` fails. If you decide to change it, there are a few more places you'll need to update this.

randomir (approved):
LGTM, just a few nits.

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1085 @@
...
+    def _canon(
+        self,
+        partition: list[set[int]],
+        trace: NDArray[np.integer],
+        color: NDArray[np.integer],
+        num_colors: int,
+        individualized_vertex: int | None = None,
+    ) -> None:
+        """Generate search tree based on iterative color refinement and vertex
+        individualization.
```
thisac, line 510: Would be nice to add to the top of the docstrings what this function and others above mutate (only `partition` I preusme; since it doesn't return anything).

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1085 @@
...
+        for u, v in edges:
+            e = (u, v) if u < v else (v, u)
+            normalized.append([tuple(label(x) for x in e)])
+        return normalized
+
+    if not all(isinstance(sublist, list) for sublist in u_vector):
+        raise ValueError("u_vector must be a list of lists.")
+
+    if not isinstance(edges, list) or not all(isinstance(e, tuple) for e in edges):
+        raise TypeError("edges must be a list of tuples")
```
thisac: Should this be above `if not u_vector`?

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1085 @@
...
+    """
+    if graph_coloring is not None:
+        coloring_keys = set(graph_coloring.keys())
+        graph_nodes = set(graph.nodes())
+        missing = graph_nodes - coloring_keys
+        extra = coloring_keys - graph_nodes
+        if missing:
+            raise ValueError(f"graph_coloring is missing nodes: {missing}")
+        if extra:
+            raise ValueError(f"graph_coloring contains nodes not in graph: {extra}")
```
thisac: It doesn't seem like `extra` and `missing` are used below? Good use for the walrus operator.
```suggestion
        if missing := graph_nodes - coloring_keys:
            raise ValueError(f"graph_coloring is missing nodes: {missing}")
        if extra := coloring_keys - graph_nodes:
            raise ValueError(f"graph_coloring contains nodes not in graph: {extra}")
```

dwave/graphs/algorithms/automorphism.py
```diff
@@ -0,0 +1,1085 @@
...
+        self._graph_coloring = graph_coloring
+        if graph_coloring is not None:
+            coloring_keys = set(graph_coloring.keys())
+            graph_nodes = set(node_order)
+            missing = graph_nodes - coloring_keys
+            extra = coloring_keys - graph_nodes
+            if missing:
+                raise ValueError(f"graph_coloring is missing nodes: {missing}")
+            if extra:
+                raise ValueError(f"graph_coloring contains nodes not in graph: {extra}")
```
thisac: Also here (if `missing` and `extra` aren't use below).
```suggestion
            if missing := graph_nodes - coloring_keys:
                raise ValueError(f"graph_coloring is missing nodes: {missing}")
            if extra := coloring_keys - graph_nodes:
                raise ValueError(f"graph_coloring contains nodes not in graph: {extra}")
```

thisac (approved):
LGTM, with also just a few nitpicks.

dwave/graphs/algorithms/automorphism.py
SebastianGitt: Thanks for the suggestion!

